PR #9, opened 2015-06-24
LTDC and DMA2D ported to ChibiOS/RT 3

TexZK:
+ LTDC and DMA2D peripheral drivers
+ LTDC and DMA2D demo project


TexZK:
In order to compile, some ChibiOS 3 core files need to be updated, as stated in the following topic:

http://forum.chibios.org/phpbb/viewtopic.php?f=16&t=2679

barthess:
Hi. Nice work. I see SDRAM related files from STDperiph lib. The community repository contains SDRAM support too but it is not tested and has no maintainer. I can not finalize it because lack of hardware and spare time. Could you finalize/test it and switch your project to it?

barthess:
Moving ili9341 driver somewhere into os/various/devices_lib will be good too.

TexZK:
Thanks! :-) This is a plain port of the old demo. I'm looking forward to help at least for the ILI9341 driver, which won't take much time.
By contrast, the SDRAM driver looks pretty complex, at least for what concerns the ST libraries, which are always a PITA to understand anyway; I assume the incomplete ChibiOS FSMCv1 driver is better engineered.

barthess:
SDRAM driver is not complex themself. But elegant mapping of its registers into human readable structure was complex task. Just take look at sources. I think it is almost done and just need testing in hardware. Man contributed this driver also wrote testhal application but he suddenly dissapeared like most of contributors.

PR #11, opened 2015-06-30
Added CRC Driver

M1Sports20:
Included in this patch is a high level and two low level drivers.

The high level driver is enabled with flag HAL_USE_CRC

The low level drivers include:
    * Hardware CRC for the STM32 cortex processor lines.(when supported)
        * Enabled with flag STM32_CRC_USE_CRC1
    * Software CRC16 and CRC32, done via lookup tables.
        * Enabled with flag CRCSW_USE_CRC1
        * CRC32 and CRC16 can be individually enabled with flags CRCSW_CR32 and CRCSW_CRC16 respectively.

### DIFF
--- a/os/hal/hal.mk
+++ b/os/hal/hal.mk
@@ -3,6 +3,7 @@ include ${CHIBIOS}/os/hal/hal.mk
 HALSRC += ${CHIBIOS}/community/os/hal/src/hal_community.c \
           ${CHIBIOS}/community/os/hal/src/nand.c \
           ${CHIBIOS}/community/os/hal/src/onewire.c \
-          ${CHIBIOS}/community/os/hal/src/eicu.c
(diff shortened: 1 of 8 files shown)


barthess:
Hi. Many times I think about writing such driver but you was first.
Code looks good but:
1) It lacks some simple testhal application. For example calculate hash from hardcoded data and compare it with software calculated hash, panic if not equal.
2) What about support of old CRC unit with unchangable polynomial (in STM32F407 for example)? Is it too ugly to support it?
3.1) #if !defined(CRC_USE_MUTUAL_EXCLUSION) check needs subcheck about CH_CFG_USE_MUTEXES flag
3.2) Code definitely needs check about suitable CRC unit presence. Also, please create patch for all stm32_registry.h files into main repository and send it via pull request.
4) please check and correct copyright notes in your files

PR #13, opened 2015-07-24
nRF51 Cleanup

doceme:
utzig has agreed to test these patches out. Please make sure he is okay with them before merging.

### DIFF
--- a/os/hal/ports/NRF51/NRF51822/nrf51.h
+++ b/os/hal/ports/NRF51/NRF51822/nrf51.h
@@ -55,6 +55,8 @@
 #ifndef NRF51_H
 #define NRF51_H
 
+#include "nrf51_bitfields.h"
+
 #ifdef __cplusplus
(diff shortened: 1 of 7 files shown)


utzig:
I would suggest some changes:

1) Please open one pull-request with your commits for bitfields and PAL updates.
2) Please open another pull-request with the serial driver updates.
3) I prefer not to merge st driver changes while I'm working on it so please don't include that commit.
4) Regarding the licensing header I would really prefer that you don't change it. I could always add your name after pulling your changes in, but refactoring per se doesn't imply updating a license (if it was, every Linux kernel file would have 100ths of owners). And of course new files should have I license header with your name.

doceme:
1 & 2) Will do.
3) No problem.
4) That's totally fine with me. I had been asked in the past to my name in the header for simple contributions, which is why I did it here, but I agree with you.  I'll only include it on new files.

PR #16, opened 2015-07-27
Cleanup nRF51 serial driver

doceme:
Use values from bitfields header. Also convert notify1 function to be
non-blocking, which seems to be how most of the other serial drivers are
implemented.

### DIFF
--- a/os/hal/ports/NRF51/NRF51822/serial_lld.c
+++ b/os/hal/ports/NRF51/NRF51822/serial_lld.c
@@ -32,9 +32,6 @@
 /* Driver local definitions.                                                 */
 /*===========================================================================*/
 
-#define INVALID_BAUDRATE 0xFFFFFFFF
-#define INVALID_PIN      0xFF
-
 /*===========================================================================*/
 /* Driver exported variables.                                                */
 /*===========================================================================*/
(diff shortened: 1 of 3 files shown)


os/hal/ports/NRF51/NRF51822/serial_lld.c
```diff
@@ -64,27 +61,67 @@ static const SerialConfig default_config = {
...
+    case 921600: speed = UART_BAUDRATE_BAUDRATE_Baud921600; break;
+    case 1000000: speed = UART_BAUDRATE_BAUDRATE_Baud1M; break;
+    default: osalDbgAssert(0, "invalid baudrate"); break;
+  };
+
+  /* Configure PINs */
+  NRF_UART0->PSELRTS = ~0;
+  NRF_UART0->PSELCTS = ~0;
+  if (config->tx_pin != NRF51_SERIAL_INVALID_PIN) {
+    // Should we assume the mode is set by the board config?
```
utzig: If the user has to add the pins to the config anyway you can configured them here so please remove these comments.

doceme: It's possible that the user didn't set a config at all and instead used the
default config because they didn't realize they needed to.  I think the
assert above would handle that case, though, so I'll remove the comment
here.

Also, I just realized that some users may want to just transmit or just
receive and may not need to use both TX and RX pins.  I'll push an update
that makes it mandatory to configure TX or RX (or both), but assert when
neither pin is configured.

PR #21, opened 2015-08-09
Add EXT driver for nRF51

doceme:


utzig:
Well done sir! Btw, IMO it would be more "idiomatic" to just get rid of ext_lld_isr.c and add the code directly to ext_lld.c.

doceme:
Thanks! Yes, I just saw another EXT driver have separate files, so I just did it based on that example. I can certainly combine them.

PR #22, opened 2015-08-16
Add CRC Driver

M1Sports20:
This patch includes a high level and two low level drivers.

The high level driver is enabled with flag HAL_USE_CRC

The low level drivers include:
    * Hardware CRC for the STM32 cortex processor lines.(when supported)
        * Enabled with flag STM32_CRC_USE_CRC1
        * DMA is enabled with CRC_USE_DMA
          * SYNC api will use DMA, but put calling thread to sleep
          * ASYNC api enabled.
        * DMA Disabled
          * SYNC api spin while calculating CRC
          * ASYNC api disabled
    * Software CRC (3 modes)
        * CRCSW_CRC32_TABLE - Enables crc32 with lookup table.
        * CRCSW_CRC16_TABLE - Enables crc16 with lookup tables.
        * CRCSW_PROGRAMMBLE - Enables any crc done with computation.
          * Can calculate any crc configuration.
        * CRC_USE_DMA obviously not support with software CRC


barthess:
1) hal_community.h. CRC is not complex driver. Complex drivers for example:
    MMC over SPI, 1-wire over PWM, 1-wire over UART, etc.
2) crc.h lines 75-76. Comment looks misplaced.
3) crc_lld.c. Strange functions _crc_lld_calc_byte() _crc_lld_calc_halfword(). 
    Is pointer saving to temporal variable really needed? Why not write
    directly to DR like in _crc_lld_calc_word()?
4) crc_lld_calc(). Your work around hardware errata. It may cause unaligned 
    memory access. As a result MPU will kill your program using (unhandled) exception.
5) crcCalc() function performs (potentially very long) calculation inside lock.
    Note: async functions need locks because of thread suspend calls but this
    is not such case.
6) Check your doxy comments in crcCalc() crcStartCalc() etc. about configured
    callback. It looks like copy-paste error.
7) Both crcStartCalcI() and crcStartCalc() have identical (redundant) checks.
8) crcsw.c. Function reflect() formatted using tabs instead of spaces.

M1Sports20:
Thank you for the review,  I have put my comments inline below.  I will
submit a follow up patch to this pull request.

On Sun, Aug 16, 2015 at 09:15:14AM -0700, Uladzimir Pylinski wrote:
> 1) hal_community.h. CRC is not complex driver. Complex drivers for example:
>     MMC over SPI, 1-wire over PWM, 1-wire over UART, etc.
Sorry, I didn't mean complex as how your describing it(second layer
driver).  I meant complex as in it has more lines of code that I would
like.
> 2) crc.h lines 75-76. Comment looks misplaced.
I will fix.
> 3) crc_lld.c. Strange functions _crc_lld_calc_byte() _crc_lld_calc_halfword(). 
>     Is pointer saving to temporal variable really needed? Why not write
>     directly to DR like in _crc_lld_calc_word()?
If you do not typecast the bus will actually do a 32-bit transaction.
So for 16 bits instead of calculating the crc for 0xFFFF it would do
0x0000FFFF if you don't typecast it.
> 4) crc_lld_calc(). Your work around hardware errata. It may cause unaligned 
>     memory access. As a result MPU will kill your program using (unhandled) exception.
That is a good idea and I didn't think about that.  However, I am not taking an
exception.  The CRC result is just simply wrong in this setup.  I will
do some more internet searching to see if anyone else has seen this.
> 5) crcCalc() function performs (potentially very long) calculation inside lock.
>     Note: async functions need locks because of thread suspend calls but this
>     is not such case.
Good catch and I don't know know why I did this.  I suspect you are right and I
will look over the code and submit a follow up patch.

> 6) Check your doxy comments in crcCalc() crcStartCalc() etc. about configured
>     callback. It looks like copy-paste error.
Will fix, thanks.
> 7) Both crcStartCalcI() and crcStartCalc() have identical (redundant) checks.
> 8) crcsw.c. Function reflect() formatted using tabs instead of spaces.
Will fix, thanks.

M1Sports20:
Everything has been updated.

Note on:
 5) crcCalc() function performs (potentially very long) calculation inside lock.
     Note: async functions need locks because of thread suspend calls but this
     is not such case.
I left lock there, but wrapped in the previously defined CRC_USE_DMA flag because the lock is still needed if DMA is on when calling the sync function.  The sync function(with dma enabled) still uses dma in the background, so it puts the calling thread to sleep.  This api is still sync but uses DMA.  I believe I copied this code from the I2C driver.

barthess:
5) hm... sync function using DMA. Is it faster than "manual" varialt? Any other benefits except working around hardware bug?
P.S. About bug you find. I am not surprised. I remember some epic erratas in silicon (incorrect work of ART accelerator, malfunctioning of DMA + crypto engine, close to imposible I2C without DMA, and some other I can not remember).

M1Sports20:
The sync function only uses DMA if you have enabled CRC_USE_DMA, note
this flag is disabled by default.

The only other benefit would be another thread can use the CPU while the CRC
calling thread is waiting on the DMA engine.  This would free up the CPU and
only be useful on large CRC calculates that would be longer than a context
switch.  Such as a system verifying a large range of memory(maybe off chip).
The function is still synchronise even though it is using DMA, this allows
the calling thread code simpler to write.

DMA does get around the suspect CRC hardware bug.  However, this
software patch does have a workaround that only occurs if data reflect
is not being used.  I have not measured the software impact of the
workaround.  Which limits writes to a byte at a time.

Thanks,
Michael

PR #23, opened 2015-08-20
Makefile paths rework

fpoussin:
This is a modification of how the file paths are handled.

Instead of having to create a git submodule in the main ChibiOS repo, you just have to specify its path and this repo's path.
It makes it possible to have the contrib repo located elsewhere, which is a lot more convenient.

Variables to set in the project Makefile:
*CHIBIOS = ../../../../../ChibiOS-RT*
*CHIBIOS_CONTRIB = $(CHIBIOS)/../ChibiOS-Contrib*
In this case both repos are in the same folder.

All the makefiles have been updated.


barthess:
Hi Fabien.
Your changes looks reasonable. But they will cause changes into SVN mirror too. So I decided to discuss it here http://forum.chibios.org/phpbb/viewtopic.php?f=3&t=2197&p=21946#p21946

fpoussin:
Hey, I replied to the forum thread. 

Thanks

PR #26, opened 2015-09-27
Add GPT driver for nRF51

doceme:


doceme:
I just realized that someone might use TIMER0 for GPT without realizing that it's being used for SysTick.  Since we'd like to support having SysTick run from another timer (possibly RTC sometime in the future), I need to create a #define to indicate that SysTick is using TIMER0 and #error if someone tries to use GPT with TIMER0 at the same time.  I'll add that as another commit to this change hopefully sometime today.

utzig:
Great work again Stephen!

doceme:
Happy to help. Thanks for reviewing! Hopefully these drivers will be useful
to others.

~Stephen

PR #27, opened 2016-02-01
Added missing HEAP_RAM definition

sdalu:
linker was generating a warning, and tests were failing on "Heap, allocation and fragmentation test"


sdalu:
Other boards seems also to have HEAP_RAM missing

utzig:
I guess this was recently added so maybe some other boards outside the official repo may not have been updated. I'll do a quick check tonight.

PR #29, opened 2016-02-02
hardware flow control from nRF51

sdalu:
Adding support for hardware flow control


sdalu:
Patch has been reworked.
As a side note, I'm wondering if it's really necessary to configure the pin mode, shouldn't this already  been done by `palInit` using the board configuration?

utzig:
If we know that the pin is supposed to have some specific functionality and must be configured in some specific way, better to just do it here, so if the user forgets to configure it in the pal config it will still work and otherwise it would be a lot harder to find out why it isn't working.

PR #32, opened 2016-02-05
Ease dev with flash/debug target helper

sdalu:
Adding two make file (jlink.mk and gdb.mk), to simplify development by adding to the Makefile

```
include $(CHIBIOS_CONTRIB)/os/various/jlink.mk
include $(CHIBIOS_CONTRIB)/os/various/gdb.mk

flash: all jlink-flash
debug: gdb-debug
```


utzig:
I think it would have been better to add those helpers to a `mk`directory or something similar.

sdalu:
I'll see to move it. In various/mk ?

PR #33, opened 2016-02-05
RTC for System Ticks

sdalu:
Use RTC0 has a tick source instead of TIMER0


utzig:
Wow, awesome work! I would only suggest you add the option of selecting between timer0/rtc in mcuconf.h. RTC could be the default.

sdalu:
I can make ifdef. But I am curious about what would make someone choose timer instead of rtc (except for a bug in rtc code).

sdalu:
It should be ok now.

PR #36, opened 2016-02-08
Random Number Generator driver

sdalu:


os/hal/include/rng.h
```diff
@@ -0,0 +1,136 @@
+/*
+    RNG for ChibiOS - Copyright (C) 2016 Stephane D'Alu
```
utzig: If you don't have too strong opinions on the copyright here I would suggest attributing this file to Giovanni Di Sirio because this looks like something that could be worked on to be upstream to the main ChibiOS repo and if that is to happen I think it would be very unlikely that Giovanni would accept code attributed to anyone else. But I'm not a lawyer so I'm not sure if this is something right to do...

@barthess What do you think about this?

os/hal/ports/NRF51/NRF51822/rng_lld.c
```diff
@@ -0,0 +1,152 @@
...
+  for (i = 0 ; i < n ; i++) {
+    /* Wait for byte ready
+     * It take about 677µs to generate a new byte, not sure if
+     * forcing a context switch will be a benefit
+     */
+    while (NRF_RNG->EVENTS_VALRDY == 0) {
+      /* Sleep and wakeup on ARM event (interrupt) */
+      SCB->SCR |= SCB_SCR_SEVONPEND_Msk;
+      __SEV();
+      __WFE();
```
utzig: Why two __WFE calls?

os/hal/src/rng.c
```diff
@@ -0,0 +1,182 @@
+/*
+    RNG for ChibiOS - Copyright (C) 2016 Stephane D'Alu
```
utzig: Same comment at rng.h applies here.

os/hal/ports/NRF51/NRF51822/rng_lld.c
sdalu: I got it from nordic devzone, I think it was here: https://devzone.nordicsemi.com/question/1555/how-do-you-put-the-nrf51822-chip-to-sleep/

os/hal/include/rng.h
sdalu: I'm ok with that. I'm waiting for @barthess opinion on it

utzig:
Well, cant wait forever...

barthess:
Hello guys. I have missed this event, sorry. 
About attributing this file to Giovanni. The best place to discuss it is forum topic. In general Giovanni claims to transfer copyrights to him with note in file header like "Concepts and parts of this file have been contributed by $USERNAME"

PR #37, opened 2016-02-11
Multiple drivers to include

fpoussin:
Hey guys,

Here are some more drivers to include.
- USB Host with a few device drivers
- EEprom (updated/Fixed)
- Timcap

I have tested the 2 last ones, the USB driver has been tested by other community members on F4s.
They don't all have examples in testhal yet.

Feel free to make some more tests before merging.

Thanks!


utzig:
I've read some of the of the code in this pull request and it looked good. But that is kinda of too big to really do an in depth review. I know you know what you're doing so I'm pretty sure this is good stuff, but I would really suggest that you made such big pulls into a lot of smaller pull requests for people to be able to review it thoroughly.

barthess:
I see one more problem. Donated but unmaintaned code worth nothing. Who will fix bugs and compilation errors caused by ChibiOS changes? Testhal applications are mandatory at least for compilability checks.

fpoussin:
I can take care of compilation errors for USB Host, and anything for the other drivers. (for GCC)
Right now, only the EEprom driver lacks examples, I can do that too.

Everything has been tested with 16.1 and GCC 4.8.

I think it's better to have something, even if it's not fully maintained. 
The community can fix drivers if they break later on, at least we have a starting point.

If you guys really want, I can squash/reorder some commits. (I worked on all drivers at the same time)
If that comes out clean I can split the pull requests for each driver.

utzig:
"I think it's better to have something, even if it's not fully maintained.". I agree with this. I find it better to accept that contrib repo might not be well maintained but at least has some good starting points for people which need one of those functionalities.

sdalu:
From an external point of view, I thought it was the point for having a Contrib directory, to have things  (even if not perfect) and a way to encourage people to contribute more

barthess:
I think it can be merged because:
1) @fpoussin will to maintain the code. Just split it in some smaller commits.
2) You all have no votes against it.
3) This repository was planned as some kind of code incubator before moving to main ChibiOS code base. And it is better to have it all in one place than spread around github.

P.S. @fpoussin this EEPROM driver is my very old work improved by some people. Originally driver had big test suite application. It can be used as testhal application too.

fpoussin:
I have made a simple testhal that read and writes in a loop for the EEPROM driver.
I can run some tests later on on my F4 discovery board for the USB Host driver.

I agree with all your points.
I have split all 3 drivers in 3 commits for the pull requests.

utzig:
+1 for merge.

fpoussin:
We will use pull request #40, I will make 2 more after it is done.

PR #41, opened 2016-02-16
Ensure ram is on

sdalu:
Ensure RAM bank is powered.
I guess the same should also be applied to WVSHARE?


utzig:
I can apply it do WVSHARE. Thanks!

PR #43, opened 2016-02-17
TIMCAP Driver

fpoussin:
This is a very simple input capture driver.
It can be used on all channels and you have full control over the callbacks.
I also added platform.mk for the STM32F3xx.

Original driver: https://github.com/dsigma/ChibiOS/tree/master/demos/ARMCM4-STM32F407-WAVESHARE-OPEN-407I-C-TIM_CAP/timcap


barthess:
What the differences between TIMCAP and EICU? When it better to use?

fpoussin:
For a simple measurement of period/width I would use EICU.

For anything special I would use timcap.
Like in the case you need to do measurements every "n" times/period or return 0 after a timeout.

Basically the only thing it does is call the callbacks when you get an input or timeout, you have to handle the data yourself. It's a really low level driver.

barthess:
Understand. Merged.

PR #49, opened 2016-04-04
Rename HAL files and header guards in kinetis port.

flabbergast:
... as to keep in line with the main repo.
(Filenames XXX_lld.Y -> hal_XXX_lld.Y and guards _XXX_LLD_H_ -> HAL_XXX_LLD_H_ )


os/hal/ports/KINETIS/K20x/hal_lld.h
```diff
@@ -22,8 +22,8 @@
  * @{
  */
 
-#ifndef _HAL_LLD_H_
-#define _HAL_LLD_H_
+#ifndef HAL_LLD_H_
+#define HAL_LLD_H_
```
utzig: IMO it would be better if you left out the _ at the end. But I don't mind very much!

utzig:
LGTM, thanks!

PR #51, opened 2016-04-05
Fix kinetis test thd

utzig:
This fixes build of all demos and testhal that uses the test thread.


utzig:
@flabbergast You might want to do a double check of this.

flabbergast:
Tested all the affected demos/testhals on my boards - working fine. (Too bad that shell_cmd hasn't been updated at the moment to work  with STATE_CHECK. It's quite a few changes in the main repo these days...)

utzig:
Agreed! Thanks for the review.

PR #52, opened 2016-04-08
Initial MSP430X Port.

awygle:
This port includes hal, pal, serial, and st drivers. It supports both large and small model code and data for MSP430X-class CPUs. It has only been tested on the EXP430FR5969 LaunchPad board.

I wouldn't be surprised if there were some things that needed to be fixed before merging - any feedback is welcomed.

### DIFF
--- a/demos/MSP430X/NIL-EXP430FR5969/Makefile
+++ b/demos/MSP430X/NIL-EXP430FR5969/Makefile
@@ -1,0 +1,205 @@
+##############################################################################
+# Build global options
+# NOTE: Can be overridden externally.
+#
+
+# Optimization level, can be [0, 1, 2, 3, s].
+#     0 = turn off optimization. s = optimize for size.
+#     (Note: 3 is not always the best optimization level. See avr-libc FAQ.)
+OPTIMIZE = 1
+
+# Debugging format.
+DEBUG = 
+#DEBUG = stabs
+
+# Memory/data model
+MODEL = large
+
+# Object files directory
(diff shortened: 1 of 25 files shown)


barthess:
@awygle 
I can not test your code and send any feedback, but I can grant you write access if you will be a MSP430 maintainer.

Any thoughts from other repository users/contributers?

utzig:
Is this based on the old MSP430 code or was it written from scratch?

poisson:
This was written from scratch. I referenced the old MSP430 port a bit, as well as the AVR and STM32F0 ports, to figure out what the various APIs were supposed to do.

I'm definitely willing to maintain this port.

One non-technical thing I did want to ask - since this is in Contrib I made sure all the files are Apache 2.0 licensed and changed the copyright line from Giovanni to me, is that the right thing?

awygle:
This was written from scratch. I referenced the old MSP430 port a bit, as well as the AVR and STM32F0 ports, to figure out what the various APIs were supposed to do.

I'm definitely willing to maintain this port.

One non-technical thing I did want to ask - since this is in Contrib I made sure all the files are Apache 2.0 licensed and changed the copyright line from Giovanni to me, is that the right thing?

utzig:
In contrib you can attribute everything to yourself I think. On main repo you can attribute most port specific and hal files to yourself but not core kernel files.

awygle:
@utzig That's my understanding as well. Thanks for confirming.

PR #54, opened 2016-04-09
Add MK20DX256 ld script with 8 KB bootloader

fredizzimo:


utzig:
@flabbergast Maybe with the new additions to standard rules.ld (flash0 to 7 sections) it would be a good time for us to get rid of those custom kinetis rules.ld files. What do you think?

flabbergast:
Yes, thanks for suggesting this! I didn't really like the custom rules_ files but I didn't know a better way then. I've done that here: https://github.com/flabbergast/ChibiOS-Contrib/commit/6c565f5a2f0f386bd2e4d6b18df4fb3c5eba4840

But I'm travelling now, so I'll need to test on more devices when I get home. I'll open an PR then.

PR #58, opened 2016-04-17
Added shell demo project

jstruebel:
I added the shell demo project that I've been using to test the shell enhancements that I've been working on and that Giovanni recently committed to ChibiOS trunk. It has an example shellconf.h file that can be used to enable the extra features.

### DIFF
--- a/demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
+++ b/demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
@@ -26,12 +26,12 @@ SerialConfig s0cfg = {
  */
 static THD_WORKING_AREA(waBlinkThread, 128);
 static THD_FUNCTION(BlinkThread, arg) {
+  systime_t time = 500;
 
   (void)arg;
 
   chRegSetThreadName("blinker");
   while (true) {
-    systime_t time = 500;
     palTogglePad(GPIO_LED_BLUE, PIN_LED_BLUE);
     chThdSleepMilliseconds(time);
   }
--- a/demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
+++ b/demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
@@ -1,0 +1,83 @@
+/*
+    Copyright (C) 2016 Jonathan Struebel
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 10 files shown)


demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
```diff
@@ -0,0 +1,83 @@
...
+ * Blue LED blinker thread, times are in milliseconds.
+ */
+static THD_WORKING_AREA(waBlinkThread, 128);
+static THD_FUNCTION(BlinkThread, arg) {
+
+  (void)arg;
+
+  chRegSetThreadName("blinker");
+  while (true) {
+    systime_t time = 500;
```
utzig: Although this will be optimized by the compiler, you would usually set the variable outside the loop!

demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/mcuconf.h
```diff
@@ -0,0 +1,58 @@
...
+ * HAL driver system settings.
+ */
+#if 0
+/* PEE mode - 48MHz system clock driven by (8 MHz) external crystal. */
+#define KINETIS_MCG_MODE            KINETIS_MCG_MODE_PEE
+#define KINETIS_PLLCLK_FREQUENCY    96000000UL
+#define KINETIS_SYSCLK_FREQUENCY    48000000UL
+#endif
+
+#if 1
```
utzig: Just curious but any reason why you are using FEI mode instead of PEE?

demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/main.c
jstruebel: Yes, I see that. I will correct that. I had copied this from the USB_SERIAL demo project, so I'll update that as well.

demos/KINETIS/RT-FREEDOM-KL25Z-SHELL/mcuconf.h
jstruebel: No reason in particular. It was how the project was setup when I had started with this file from @flabbergast's Kinetis branch and since it worked I didn't try to change it. I'm not familiar with the FEI mode vs PEE mode (guess it's time for me to do some reading). Is one better than the other?

flabbergast:
PEE/FEI: Probably my mistake - I've been testing all the modes that are implemented and probably just forgot that I left the "non-standard" one enabled for KL25Z. (As a curiosity - I was also testing if USB will work properly in FEI mode - which is out-of-spec - and it did.)

utzig:
I see no problem in using FEI. I was just curious if there was some reason for choosing it!

PR #62, opened 2016-04-26
[KINETIS] I2C update (timeouts, KL27Z)

flabbergast:
Slight rewrite of the I2C driver.
 - shuffled the if branches in ISR around a bit, added some checks (e.g. master/slave), and comments
 - removed S&S_BUSY waits (I haven't seen them in other kinetis I2C drivers and the datasheet also doesn't mention that it's needed - but I may be wrong about this)
 - implemented timeouts
 - fixed the base frequency for clock speed (SYS->BUS) as reported by Fred on the forums
 - added a workaround for repeated start for KL27Z MCUs (as reported [here](https://community.freescale.com/thread/377611), my own tests confirm this)

### DIFF
--- a/os/common/ext/CMSIS/KINETIS/kl25z.h
+++ b/os/common/ext/CMSIS/KINETIS/kl25z.h
@@ -777,57 +777,9 @@ typedef struct
 
 /****************************************************************/
 /*                                                              */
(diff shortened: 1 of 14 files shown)


utzig:
Nice job!

PR #68, opened 2016-05-03
MSP430X DMA Support, EXP430FR6989 Demo + Makefile Updates

awygle:
This pull request adds DMA support and a testhal test case for it, as well as a port and a demo for the EXP430FR6989 LaunchPad board and the updates to Makefiles required to remain in sync with the main ChibiOS repository. It also includes a few other housekeeping fixes such as support for GCC 6.1 and "trap" ISRs for all unused ISRs in demos (this can't be done in a graceful way yet due to compiler limitations).


awygle:
If you'd like me to split this into three separate pull requests, just let me know. Otherwise I'll give it a couple days for comments and then merge it.

barthess:
Splitting pull request suitable for code revue by other developers. But you are only developer of MSP430 port. You can try to send Giovanni link to this page but he not impressed by git/github. Any way your code is look clean and well formatted.

awygle:
Thanks for the feedback. That's approximately what I expected to hear, just trying to be a good collaborator. Maybe someday there will be another developer, eh?

utzig:
I would say that as long as you are the sole developer/user just do as much as you can by pull request and minimize the amount of work with git/github. The only exception would be if you do changes to files which are not specific to the port (can it happen at all?). In that case it would be good to do a separate PR which enables review by other people.

PR #69, opened 2016-05-05
PWM support for nRF51

sdalu:
callback only supported for now (no GPIOTE/PPI shortcuts)


sdalu:
frequency*period is off by a factor 2 (don't know why)

PR #72, opened 2016-05-08
Add qei driver

pl4nkton:


os/hal/ports/STM32/LLD/TIMv1/hal_qei_lld.h
```diff
@@ -0,0 +1,347 @@
...
+
+/**
+ * @brief   QEI counter type.
+ */
+typedef uint16_t qeicnt_t;
+
+/**
+ * @brief   QEI delta type.
+ */
+typedef int16_t qeidelta_t;
```
barthess: Not sure but it looks like potential integer overflow. Result of uint16_t - uit16_t may be more than int16_t able to store.

os/hal/src/hal_qei.c
```diff
@@ -0,0 +1,214 @@
...
+ * @param[in] qeip      pointer to the @p QEIDriver object
+ * @return              The delta from last read.
+ *
+ * @api
+ */
+qeidelta_t qeiUpdate(QEIDriver *qeip) {
+  qeicnt_t diff;
+
+  osalSysLock();
+  diff = qeiUpdateI(qeip);
```
barthess: Here you store value in unsigned variable and return them as signed. Is it correct?

barthess:
Please write your name in copyright notes in main.c files. Also you can state your name as copyright holder in HAL files.

marcoveeneman:
To me this looks very much like an updated version of the QEI driver tinito posted on the [forums](http://www.chibios.com/forum/viewtopic.php?f=8&t=685&hilit=qei) in 2012, so i think he should get the copyright of the driver.

pl4nkton:
@marcoveeneman Yes it is noted in the commit message. I place the author to the header.
@barthess Good point.

barthess:
Thanks.

PR #73, opened 2016-05-09
HAL AVR USB device driver support

rlippert:
This PR adds an interrupt-driven USB device mode driver to the AVR port similar to the existing STM32 USB implementation. No support for USB host mode or OTG at this time as I don't have any AVR chips to use that support those modes.

Includes a demo app and board definition so that the functionality can be quickly tested against a PJRC Teensy2++ board (http://pjrc.com/store/teensypp.html).

### DIFF
--- a/demos/AVR/RT-TEENSY2-USB/Makefile
+++ b/demos/AVR/RT-TEENSY2-USB/Makefile
@@ -1,0 +1,568 @@
+# Hey Emacs, this is a -*- makefile -*-
+#----------------------------------------------------------------------------
+# WinAVR Makefile Template written by Eric B. Weddington, Jörg Wunsch, et al.
+#
+# Released to the Public Domain
+#
+# Additional material for this makefile was written by:
+# Peter Fleury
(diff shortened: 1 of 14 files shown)


utzig:
Wouldn't it be better to contribute this to the official repo since the AVR port is there?

rlippert:
I originally tried that but it got rejected: https://github.com/ChibiOS/ChibiOS/pull/11

I think there are copyright and support issues with the official repo, the owner seems to only like to submit self-created code or small fixes from community.

utzig:
Make a patch with the same contents as this pull request against the main repo and I will apply it. Going through github doesn't really make any sense for the main repo because it's just a read-only mirror.

rlippert:
I've attached a patchset reworked against the main SVN codebase as a gist here: https://gist.github.com/rlippert/f6ab7b56a8e4e6e3eb094f340970bee9

utzig:
This was applied against the main repo so I'm closing here. Great work btw!

PR #81, opened 2016-06-05
MSP430X ADC12 Support

awygle:
Adds support for the MSP430X's 12-bit ADC peripheral, as well as
reasonably complete testing of same.

Also includes fixes for several bugs and cleanup of the DMA peripheral,
which used ch calls rather than osal calls and was unclear about what
contexts its methods could be called from.


utzig:
Instead of having to add an msp_vectors.c file to every demo, can't you just add a vectors.c with all vectors declared as weak like in the ARMCMx port?

PR looks good to me.

awygle:
I would *love* to do that, unfortunately as far as I can tell the GCC compiler for MSP430 doesn't currently handle weak interrupts correctly (see my GCC bug report at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70713 which has received zero attention).

barthess:
May be it worth to create simple script (python, perl, etc.) for msp_vectors.c generating from halconf.h. Halconf.h has regular structure and can be easily parsed using regular expressions.

PR #85, opened 2016-06-28
QEI driver for NRF51

sdalu:
Quadrature decoder for NRF51.
* Not tested with led pin (I don't have the necessary hardware)
* should I define a type for overflow_cb? or use the qeicallback_t?
* behaviour of counter overflow is user controllable
* trying to introduce new driver methods qeiSetCount and qeiAdjust (should it be in qei implementation?)
* are callback correctly handled? (they are supposed to be called from interrupts, or threads)

Comments and feedback are welcomed.

### DIFF
--- a/os/hal/boards/NRF51-DK/board.h
+++ b/os/hal/boards/NRF51-DK/board.h
@@ -44,6 +44,16 @@
 #define SPI_SS         24
 #define I2C_SCL         7
 #define I2C_SDA        30
+#define AIN0           26
+#define AIN1           27
+#define AIN2            1
+#define AIN3            2
+#define AIN4            3
+#define AIN5            4
+#define AIN6            5
+#define AIN7            6
+#define AREF0           0
+#define AREF1           6
 
 /*
  * IO pins assignments.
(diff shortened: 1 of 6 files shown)


barthess:
@sdalu do you think your pull request finished? I can find some spare time to review it.

sdalu:
@barthess, yes pull request is finished.
I've some design questions (mentioned above), that you will surely be able to answer during your review.

barthess:
> should I define a type for overflow_cb? or use the qeicallback_t?
> behaviour of counter overflow is user controllable

I think there is no need of overflow_cb because all overflow logic handled under the driver's hood.

>trying to introduce new driver methods qeiSetCount and qeiAdjust (should it be in qei implementation?)

I think it is better to handle adjustment in user level code. Are they really needed?

> are callback correctly handled? (they are supposed to be called from interrupts, or threads)

Callbacks generally use to notify threads, not to call from them. If I correctly understand your question.

sdalu:
When I wrote the code I had a rotary button in mind:
* the idea of the overflow_cb is to know that we've reached the "end" (for the non-wrapping cases) and possibly notify user of it (which can't be done with counter-changed callback as value doesn't change)
* qeiSetCount / qeiAdjust is for external adjustement of the rotary or rotary which have a position 0 encoded

Perhaps I'm trying to put too much in the driver.

Removing qeiSetCount/qeiAdjust would avoid having a callback being called from thread level (which I undestand is not a nice way to use callback).
But not being able to set/adjust driver counter from user code, I fell then that the overflow logic and even qeiGetCount would lose usefulness as user would need to reimplement it from the delta value to manage the counter himself. That's my dilemma

A more lightway driver could be to just provide the accumulated delta value and the generic qei_adjust_count function as an helper, but the SMT32 implementation is counter-based and I don't know if it is possible to have it working the other way arround

barthess:
Ok. Lets do the following. Move all your hal_qei related code to its destination (hal_qei.c/h). I will merge code. Later problems will be solved in future.

utzig:
I agree with barthess that is best to just merge as is and let fixes come later.

Btw, @sdalu you should avoid `//`style comments for consistency with ChibiOS/RT style.

sdalu:
Thanks for the merge.
@barthess, I forgot one part, so I created a new pull request https://github.com/ChibiOS/ChibiOS-Contrib/pull/87 I let you see if it is really useful to merge it now or not.

PR #89, opened 2016-07-08
Making room for NRF52832

sdalu:
Renamed nrf51 directory. Upgraded headers to SDK11


utzig:
What do you think of `NRF5x` instead of `NRF5`?

sdalu:
Nordic choosed NRF5 as a unified name for its software
http://developer.nordicsemi.com/nRF5_SDK/doc/

If you think NRF5x is a better naming for the hardware, I don't have a strong opinion about it.
Let me now and I will change it.

PR #91, opened 2016-07-09
Nrf52832

sdalu:
First implementation of board NRF52-DK, and port NRF52

* Renamed NRF51_* constantes to NRF5_*
* One test case failure: "out of time window", but that's ok
* Implemented drivers: pal, st, uart, rng, wdg, gpt, qei
* Converted but not tested: spi, i2c

Need consideration, but can be done later:
* Is it still useful to keep TIMER0 as a system ticks implementation?
* peripheric configuration should use new ioline_t type instead of pad? (with nrf52 release looks like more ioports could be in consideration for future series)
* freerunning doesn't compile (but was already broken on nrf51 due to osal not included)


sdalu:
Test case failure is just by 1 tick, due to '<' in time_window assertion, I asked forum to see if this a real problem.
http://www.chibios.com/forum/viewtopic.php?f=25&t=3446&p=26058#p26058

sdalu:
Hi @utzig can you have a look? Thanks

utzig:
@sdalu I will take a look tomorrow.

PR #95, opened 2016-08-30
set sampling channel PWM output to disabled so the IO pin is free

tixiv:
Tested with DS18S20 temperature sensor on STM32F072, working fine.


barthess:
Could you provide more descriptive comment? What problem solves your patch?

PR #98, opened 2016-11-06
STM32 CRC : Fix asserts

kimmoli:
These asserts will trigger when !(config != default_config), and IMHO just reverse is expeceted when CRC is not programmable.


barthess:
Looks like CRC contributor is no more with us. I just close my eyes and press "merge".

PR #101, opened 2016-11-15
Fix KINETIS USB_SERIAL testhal compilation.

funnydog:
Fix KINETIS USB_SERIAL testhal compilation by renaming the function
sduDisconnectI() to sduSuspendHookI().

### DIFF
--- a/testhal/KINETIS/FRDM-K20D50M/USB_SERIAL/usbcfg.c
+++ b/testhal/KINETIS/FRDM-K20D50M/USB_SERIAL/usbcfg.c
@@ -284,7 +284,7 @@ static void usb_event(USBDriver *usbp, usbevent_t event) {
     chSysLockFromISR();
 
     /* Disconnection event on suspend.*/
-    sduDisconnectI(&SDU1);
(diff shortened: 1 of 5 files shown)


utzig:
Thanks. Maybe adding the `sduWakeupHookI` (plus locking) to `USB_EVENT_WAKEUP` would be also be a good idea.

funnydog:
Hello,

I prepared a patch to address the missing sduWakeupHookI() calls for every KINETIS boards.

While doing that I realized that USB_EVENT_RESET and USB_EVENT_UNCONFIGURED were handled incorrectly compared to the code of the STM32 targets. 

Therefore I changed also the switch statement to have the same logic as the STM32 targets.

However I'm unsure about testhal/KINETIS/FRDM-KL25Z/USB_HID/usbcfg.c: the USB_EVENT_SUSPEND is not wired to sduSuspendHookI(). Maybe it's that way on purpose?

As soon as I test the changes on my Teensy 3.2 I'll send a new PR.

funnydog:
never mind the comment on USB_HID, there is no SDU in that case.

PR #103, opened 2016-11-18
Simple radio support with NRF51 implementation

sdalu:
Feedback welcome (not to be merged yet)


awygle:
This seems to me like something that should be an EX interface rather than being added to the community HAL, if you're trying to make it be a generally useful interface to RF chips. The NRF is built in obviously but in most other cases the RF chips will be external.

sdalu:
I would say more like the MAC driver which is part of HAL

awygle:
I don't know too much about the MAC driver but I don't know if off-board discrete MAC chips are very common (I'm more used to seeing MAC+PHY together). While off-board RF chips attached by SPI are very common and it would be nice to be able to use the existing SPI drivers for them while still following this interface.

fpoussin:
Closing this on as it's too old for current code base and there are alternatives.

Issue #84, opened 2016-06-25
qeiUpdate, unexepcted value due to counter wrapping


sdalu:
In the QEI driver, the qeiUpdate function returns the counter delta from the last reading, which when the counter is wrapping result in a huge delta.

| prev count | current count | delta | expected delta |
| --- | --- | --- | --- |
| 0 | 65535 | 65535 | -1 |
| 65535 | 0 | -65535 | +1 |

barthess:
I am not an author of QEI driver but I can review your pull request and merge it.

sdalu:
The fix looks like reverting commit https://github.com/ChibiOS/ChibiOS-Contrib/commit/4a36fb944b854eb91e74ce18a7b2cb0cedf7d0d1 . Initial code was relying on the wraping of the int16_t when substracting two uint16_t to provide the correct value. Which still open a can of worm:
- overflow of signed int is undefined behaviour in C/C++
- writting new driver impose to chose identical size for qeidelta_t and qeicnt_t and no mention of it is made
- value will still be wrong if delta is really above 32767 (int16_t)

I think qeiUpdate methods should be removed:
- values returned by qeiUpdate can be meaningless if counter has wrapped several times between requests
- delta values would be better provided by a driver callback when updating the counter, which will also make size of qeidelta_t and qeicnt_t independant. (possible with nRF51, but I don't know enough about STM32)
- if driver is not able to provide such a callback, user can still take the risk to implement its own version of current_value - last_value

fpoussin:
I merged a pull request that reverts the said commit. (among other things)

PR #107, opened 2016-12-05
Stm32 fixes

pl4nkton:


fpoussin:
Can you add a check here: https://github.com/ChibiOS/ChibiOS-Contrib/pull/107/files#diff-e86bb84ff7328e855b605b232a081efeR1390 ?
In case the _STM32_OTG_STEPPING_ define is missing, we want to have a compiler error.
I assume this define would be located in _mcuconf_community.h_, don't forget to add it to _testhal/STM32/STM32F4xx/USB_HOST/mcuconf_community.h_

pl4nkton:
The STM32_OTG_STEPPING is provided by the Chibios/RT HAL device registry.

fpoussin:
Ok, are the QEI changes related to #84 ?

pl4nkton:
Yes. But like sdalu comments in #84 some problems remain.

PR #112, opened 2017-02-08
Comparator Driver

fpoussin:
This is a driver for the Comparator peripherals.

I need to do some testing but you can already look at the code.


barthess:
It looks clear and simple for my taste.

fpoussin:
Tested and working on F3 discovery board.

PR #115, opened 2017-04-20
Various tiva improvements

marcoveeneman:
This mainly adds advanced buffering support for the tiva serial driver.

Note that an initial .travis.yml file has been added to test if the Tiva projects succeed in building. To use this you have to enable [Travis](https://travis-ci.org) access to your repository.

If no one has any comments i will merge this next Monday (24th).


fpoussin:
Can you remove the Travis part or at least move the scripts to /tools?
We already have CI, I can put the build scripts here if needed.

Otherwise this looks good.

marcoveeneman:
Yes of course.

I didn't know about the already existing CI, what are we currently using? Adding the build scripts here would be very useful, at least for me. I didn't see any so i assumed CI was not used yet.

fpoussin:
I have a Jenkins instance running a build every 6 hours (I can't create hooks), we can still add Travis in top of that, that would work if we use the same build scripts.
You can see the results in the commits list.

fpoussin:
I added the build script if you want to have a look.

marcoveeneman:
Ah i see it now, nice. I see it is totally different from how Travis works.

I can't move the .travis.yml file to the tools directory (if we want to keep it) because it has to live in the project root to work :(

We can do both Travis and Jenkins, i think. But one thing I like more about Travis compared to your Jenkins instance is that, IMO, it is better integrated with GitHub. All pull requests to the base are automatically tested and forks get CI by just enabling it in their own configuration if they want. If not, they can just leave it disabled.

So, how will we proceed?

fpoussin:
We can keep the travis file in root and move the scripts to /tools.

Indeed we can do both so that's what we'll do.
With travis you still need admin rights to create the hooks, jenkins also does it automatically but it's more setup work as you need tokens. I can't enable PR build test in jenkins as I don't have these here...

fpoussin:
All good, thanks!

PR #116, opened 2017-06-05
Add checks to STM32 QEI driver if TIMx is already used

andresv:
For example PWM driver has the same checks: https://github.com/ChibiOS/ChibiOS/blob/master/os/hal/ports/STM32/LLD/TIMv1/hal_pwm_lld.h#L251-L306.


fpoussin:
All good, thanks!

PR #117, opened 2017-06-06
USB Host update

dismirlian:
Hi,

Here is an update of the USB Host driver:
- Implemented new driver: AOA (Android Accessory)
- Implemented new driver: HID (basic function only; no parsing of HID report descriptors)
- Implemented new driver: UVC (USB Video Class)
- Reworked MSD (Mass Storage Driver) to include error handling as per the USB spec.
- Fixes to FTDI driver
- Lots of fixes/cleanup for the main host driver
- Fixes to the LLD STM32 host driver
- Updated example/test application (for F407 Discovery board)

Some comments on the new drivers:
- AOA: allows ChibiOS to act as master to an Android device. Android can automatically launch an app when it detects the ChibiOS host, and opens a "virtual serial port" with ChibiOS.
- HID: currently, only the "boot protocol" is supported. All standard mice and keyboards support it, but special keys are generally not reported. The driver provides callbacks which inform the main application that a key was pressed, or a mouse was moved, but no parsing is done on the reports (for now, this is up to the main application; any contributions welcome!).
- UVC: basic driver for webcams (video only, no audio!). The test application shows an example of ChibiOS grabbing frames from a WebCam video stream, and saving them to an USB stick in JPEG format (both connected through a USB HUB to the STM32).

Tested with ChibiOS 16.1.8, with OTG_FS, on a custom board (STM32F411@100MHz).

This is my first PR ever; I'm not sure I'm doint it right!

Thanks,
Diego.


fpoussin:
Excellent work, thank you!

dismirlian:
Thanks Fabien!

Do you think that I should target the development version of ChibiOS (eg. RT4) instead of the stable version (eg. 16.1.8)?

fpoussin:
I have fixed a couple things to make it work on RT4.
I think that's better to target it now since it's going to be stable soon.

dismirlian:
Yes, I saw your fixes; that's why I asked. I didn't know that RT4 was coming soon. For the next commits, I'll target RT4.

Thanks!
Diego.

PR #133, opened 2017-11-04
Fix for K20x startup

belak:
While working on QMK support for the K-Type keyboard (which uses this code) I ran into an issue where it would only start up about a third of the time. It was hanging on this:

```
   /* Wait for the MCGOUTCLK to use the oscillator */¬
   while ((MCG->S & MCG_S_CLKST_MASK) != MCG_S_CLKST(2));¬
```

I stumbled on this solution by accident but it seems to fix the issue by setting the C2 register using only one assignment, similar to how they do it in the reference manual.

https://github.com/qmk/qmk_firmware/pull/1963


utzig:
Nice, thanks!

PR #137, opened 2017-12-14
Add STM32F769 to FSMCv1 sdram driver

dflogeras:
Tested on STM32F769I-Discovery DISC1 board.


barthess:
Merged. Thanks!

dflogeras:
barthess, I just looked at this patch again (setting up another environment), and I wonder if I shouldn't have changed hal_fsmc.h.

I changed a generic STM32F7 to STM32F769 (not sure why I did this), but it seems that might not be correct?

barthess:
It's better to have detailed define instead of generic one.

PR #141, opened 2018-02-03
Add usb_lld_wakeup_host for Kinetis

fredizzimo:
This make it compatible with the 17.6.3 version, and maybe some earlier ones as well.


fpoussin:
LGTM

PR #146, opened 2018-03-11
Keep track of STM32 RCC API

romainreignier:
RCC API changed in 01/2018 so apply the changes.

Note that ae7a4d40b84d8afc999691577210696f16e682f6 partially fixed the changes in QEI module but some were missing.
So update the other modules too.


fpoussin:
I was going to do just that, thanks!

fpoussin:
I have some STM32 boards for F0/3/4.
we need to make the all compile first.

PR #148, opened 2018-03-12
platform: add support for STM32L4 family

romainreignier:
This file allows to build for STM32L4xx platform.

I have only tested the QEI module.

So, do I have to only add this module or all of them and drop some if a user report an error?


fpoussin:
Thanks!

PR #150, opened 2018-03-16
Changed directory structure of Tiva PAL testhal

marcoveeneman:
Changed directory structure of Tiva PAL testhal and fixed some small compile warnings


fpoussin:
All good, thanks!

PR #152, opened 2018-03-21
WIP: Update to new Makefiles format

ycherniavskyi:
Do anyone interested in updating to new Makefiles format of ChibiOS where shared variables (such as `ALLCSRC` and `ALLINC`) and smart build used?

If yes then I could continue to developt this pull request. But it definitely breaks the current build system of ChibiOS-Contrib users.

For example after the first commit of this PR all Makefile must be updated and instead
```
include $(CHIBIOS_CONTRIB)/os/hal/hal.mk
```
use
```
include $(CHIBIOS)/os/hal/hal.mk
include $(CHIBIOS_CONTRIB)/os/hal/hal.mk
```
(which for me persomaly looks more obvious).

### DIFF
--- a/os/hal/hal.mk
+++ b/os/hal/hal.mk
@@ -1,27 +1,31 @@
-include ${CHIBIOS}/os/hal/hal.mk
+# List of all the ChibiOS/HAL community files
+HALCONTRIBSRC := ${CHIBIOS_CONTRIB}/os/hal/src/hal_community.c \
+                 ${CHIBIOS_CONTRIB}/os/hal/src/hal_nand.c \
+                 ${CHIBIOS_CONTRIB}/os/hal/src/hal_onewire.c \
+                 ${CHIBIOS_CONTRIB}/os/hal/src/hal_eicu.c \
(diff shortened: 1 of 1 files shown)


ycherniavsky:
@utzig @fpoussin could you please clarify your view on updating current ChibiOS-Contrib build system, especially the part about removing `include ${CHIBIOS}/os/hal/hal.mk` from `os/hal/hal.mk`?

I need to know the viewpoint of this repository maintainers before continue developing this MR.

os/hal/hal.mk
```diff
@@ -1,27 +1,31 @@
-include ${CHIBIOS}/os/hal/hal.mk
```
fpoussin: We need this back

fpoussin:
Hi,
You can implement it, that would be great! :)
We would like to have as few modifications as possible on project (demos) Makefiles as possible compared to vanilla ChibiOS, so you'll have to include some of the original mk files. You also need to create the driver.mk files.

fpoussin:
Replaced by #167

PR #153, opened 2018-03-22
PID Library and demo

fpoussin:
I ported the Arduino PID library for our use.


fpoussin:
I'll merge as soon as I'm done tweaking the example.

PR #157, opened 2018-05-01
Fix compilation errors with latest code from ChibiOS master branch

walkerstop:


walkerstop:
Kinesis MCUs cannot build with latest ChibiOS.  A few updates are required.

fredizzimo:
There are some duplication here, parts of these changes were already added in https://github.com/ChibiOS/ChibiOS-Contrib/pull/141

@walkerstop, since you are doing changes for QMK, we need to discuss on the Gitter channel about this, because it's likely that you are doing something wrong, the version we are using, should already have the changes from #141 integrated.

walkerstop:
Hi @fredizzimo , yes I had made some mistake syncing my fork with the upstream so I guess my fork didn't have the chnages from #141 
I synced the fork and I think the pull request is correct now

PR #165, opened 2018-07-12
Fix #161: STM32 LLD CRCv1 large data bug in DMA mode

### LINKED ISSUE #161: [CRC] Max size bound to max single DMA transfer
Hi,

there is a problem with CRC Driver (community) with CRC_USE_DMA=TRUE,
driver does not work correctly with buffer more 256k due to DMA limitation. 
ChibiOS version - 17.6.3

lechndo:
Hi guys,

sorry for my issue duplicate.

As you already mentions correctly in #161 we have to split up the DMA transfer because the STM32 DMA can only handle 65535 bytes per transfer.
This is my current fix and the results are very promising.

What do you think about my solution?


fpoussin:
LGTM

PR #166, opened 2018-08-06
Fix incompatibilities between ChibiOS head and USB Host driver

MegabytePhreak:
I was trying to use the USB Host driver with the latest ChibiOS and ran into some issues both build time and runtime. These are the fixes. The VMT updates are not 100% complete as they do not include the (new?) ctl method for the aoa and ftdi drivers. I'm not inclined to figure them out since I won't be using either driver.


fpoussin:
I don't think we need this after the last merge, can you confirm? (except the ftdi fix)

fpoussin:
Okay, then closing this PR.
Thanks!

PR #169, opened 2018-10-01
STM32 USBH LLD Fixes

apmorton:
As per commit messages:
- implement support for ISO/INT out transfers
- some minor compiler warning fixes related to debug log levels
- fix a potential infinite loop when closing an endpoint


fpoussin:
Thank you!

PR #170, opened 2018-10-16
Added setting qei to value for stm32

kajusK:
The qeiSetCount prototype was not in header file (causing -Wimplicit-function-declaration warning in gcc) plus the stm32 port had empty body of qei_lld_set_count.


fpoussin:
Thanks!

PR #171, opened 2018-11-01
Fix for rccEnableCRC macro parameter.

dflogeras:
This was removed in commit
ae7a4d40b84d8afc999691577210696f16e682f6#diff-7ddaa5ecc31109f41b7801dea2660b47

But I think is still necessary as the underlying rccEnableAHB macros
take parameter 'lp'.  It seems to work for the F0xx series, because its
rccEnableAHB ignores the 'lp'.  It is required when I tried to use the
CRC driver on a family that does require the 'lp' parameter in the lower
level macros.


fpoussin:
I must have missed that, thanks!

PR #175, opened 2018-12-10
Fixed board generation for pins with note in name

kajusK:
Some pins are ignored when the pin name in stm32cubemx xml is in format like:
PC14-OSC32_IN
PH0 - OSC_IN

This can cause build errors (e.g. for stm32f412 the VAL_GPIO_X are not defined in generated output and the build fails when the board.c tries to use these defines).

This patch fixes the port name generation.


fpoussin:
Thank you!

PR #176, opened 2018-12-21
Fixed Timer number generation

kajusK:
S_TIM will never appear in the signal list as the S_ is removed in read_project(), I guess the required output is the timer number instead of text TIMx where x is the timer number. It won't fail during the compilation as TIMx is usually defined in CMSIS headers, but...


kajusK:
Included in #177, this request can be ignored

PR #177, opened 2018-12-21
Generate peripheral and channel numbers from signal names

kajusK:
Added code for generation of peripheral and channel numbers. Eg. for ADC generate adc number and adc channel for each signal so it could be used in code. Similar to already existing generators for timers.


fpoussin:
I had not updated this for a long time, thank you!

PR #178, opened 2019-01-08
added NRF52 pwm, icu, i2c, radio esb drivers

AndruPol:
Hi !

i tested code with ChibiOS 18.2.x and it compiles without problems


fpoussin:
LGTM, thanks!

PR #181, opened 2019-01-28
Added ODR setting to mx2board.py

kajusK:
Hi, the mx2board script was not setting the the pins state for outputs correctly, by default the stm32cube sets pins to low, the default state in the script is high. Also the PinState value was not considered when generating the output.


fpoussin:
Great, thank you!

PR #183, opened 2019-02-10
Added spi and fixed FSMC alternate setup

kajusK:
Hi,
I've added generator of SPI device in addition to others like I2C, etc.

Also I've noticed that the script fails to generate FSMC AF registers correctly as the name of the signal is FSMC_D12_DA12, but there are two different signals in the gpios - FSMC_D12 and FSMC_DA12, at least for the STM32F412. I'm not really familiar with the different types, but I guess that's because these signals share the AFRx register values.


fpoussin:
Thanks!

PR #185, opened 2019-03-06
Fixed GPIO_PIN_RESET instead of GPIO_PIN_CLEAR

kajusK:
The latest stm32cube project file is crashing the mx2board.py - output low pin state is named GPIO_PIN_RESET instead of GPIO_PIN_CLEAR.


fpoussin:
Thanks!

PR #186, opened 2019-03-07
Add call to cache flush on disk write

josesimoes:
- This is required for F7 targets (and others that feature memory cache).
- The call is provided empty for all the other series, so it's OK to use it as it is.


fpoussin:
LGTM

fpoussin:
https://github.com/ChibiOS/ChibiOS/blob/master/os/hal/ports/common/ARMCMx/cache.h#L110

fpoussin:
Thanks!

josesimoes:
You're welcome. 👍 

I'm anticipating that this will be added to ChibiOS repo code following [this](http://www.chibios.com/forum/viewtopic.php?f=16&t=5025) forum thread .

PR #187, opened 2019-03-10
Fix F7 cache invalidation on write operation

josesimoes:
- This should be the correct fix.


fpoussin:
Was this tested?

josesimoes:
It builds OK.
I've tested separately on a system with an sdio SD card and another with an USB MSD.

fpoussin:
Thanks!

PR #190, opened 2019-03-13
I2C not set as opendrain

kajusK:
The I2C bus was not set to opendrain output type which caused very hard to find bugs.


fpoussin:
Thanks!

PR #191, opened 2019-04-09
NRF5x: Add support for PAL events.

ObKo:
Remove obsolete EXT driver and add support for events in PAL driver - same as in STM32 PAL driver.

NRF5x has 8 (4 for NRF51) channels for GPIO event handling. Each pad can be configured to generate one of those events. This PR introduced special macro `NRF5_PAL_PAD_TO_EVENT(pad)` that makes pad->event mapping. By default `event = (pad % 8)`, and users can redefine that macro for own purposes.


fpoussin:
Build failed, please fix.

ObKo:
The problem was in new ICU driver - GPIOTE interrupt was used exclusively by it somehow.

fpoussin:
Thank you.

PR #198, opened 2019-10-02
Fault handlers

dismirlian:
This PR adds a generic fault handler mechanism, particularly for ARMv7m (Cortex M3/M4(F)/M7).

The fault handlers include:
- HardFault handler, which decodes the fault registers and generates a description string.
- _unhandled_exception handler, which decodes the active vector, so you can effectively see which interrupt was triggered.

The HardFault handler [(here)](https://github.com/dismirlian/ChibiOS-Contrib/blob/1ca49e0b903954c1fd463d29e51faa22436fa33f/os/common/ports/ARMCMx/compilers/GCC/utils/hardfault_handler_v7m.S#L38) does the following:
- Switches to a separate stack (in case of a corrupt SP).
- Decodes the hard fault registers (done in C).
- Calls an application-defined hook
- Creates an artificial stack frame for the debugger and causes a breakpoint, in order for GDB to show a meaningful stack unwind, so you can see the whereabouts of the instruction that caused the fault.
- If no debugger is attached, restores the separate stack and calls chSysHalt("hard fault")

The _unhandled_exception handler:
- decodes the active vector, so you can effectively see which interrupt was triggered.
- If no debugger is attached, calls chSysHalt("unhandled exception")

Also, I included a demo project. Any suggestions are welcome!

Cheers,
D.


fpoussin:
LGTM, merging.

PR #200, opened 2019-10-03
SEGGER RTT/SystemView bindings

dismirlian:
This PR adds bindings for SEGGER RTT and SystemView.

You will find an example project in demos\STM32\RT-STM32F407-DISCOVERY-RTT+SystemView.


dismirlian:
Hi, CI build fails because Jenkins will need to decompress the (new) `SEGGER-RTT_6.44i+SystemView_2.52h.7z` file in `ChibiOS-Contrib\ext\`.

@fpoussin Do you want me to try to fix it? I don't have too much experience with Jenkins, but I think I know where and how to implement the fix.

PR #206, opened 2019-11-03
+ Added board files and example for STM32F411-Discovery board

WowSuchName:
Not sure whether this is perfect or not, but it worked well for me.


fpoussin:
Thanks!

PR #209, opened 2020-01-05
Update NRF52832 device headers to latest version from NRF5 SDK

ObKo:
... so they will be compatible with SoftDevice BLE stack headers.


fpoussin:
Thank you!

PR #212, opened 2020-01-12
Driver for new ADC peripheral (called SAADC) in NRF52 devices.

ObKo:
Checked with provided sample (on other custom board though).


fpoussin:
Looks good, thank you!

PR #216, opened 2020-04-09
add support for STM32H7

pierstitus:
add makefile for stm32h7 devices and add USB_MSD example for nucleo-H743

Only USB_MSD is tested and works fine, other parts are not tested.


fpoussin (approved):
LGTM

PR #223, opened 2020-09-21
Added comment that might help others to prevent cache coherence issues.

markuspetermann:
Knowing this may save time on debugging.

However, as the (default) .nocache section is rather small, a better solution would be to include cache flushing / invalidation into the read and write functions.

See: http://www.chibios.org/dokuwiki/doku.php?id=chibios:articles:cache_coherence


fpoussin (approved):

fpoussin:
Thank you!

PR #232, opened 2020-12-02
[HAL] [NUC123SD4AN0] FIX: NUC123SD4AN0 does not build

alexclewontin:
Hello,

I have been working on a port of the HAL for the Nuvoton NuMicro NUC123 series, and I just saw that some drivers for the platform were recently committed here, which is great. I started trying to integrate my work with what was in the repository, but I pretty quickly ran into some issues, and before I continue on trying to integrate the various peripheral drivers, I want to make sure the base platform (clock distribution, etc) is sorted out.

As they currently exist, the NUC123SD4AN0 drivers do not build (at least with gcc). The first problem is the platform.mk file doesn't work, and trying to build hal.c results in 

```
.../ChibiOS/ChibiOS/os/hal/include/hal.h:122:10: fatal error: hal_lld.h: No such file or directory
  122 | #include "hal_lld.h"
      |          ^~~~~~~~~~~
compilation terminated.
```
Once the platform.mk file is modified to successfully locate the compiler produces several hundred warnings about undefined macros. These are from undefined (and unused) compile-time settings (some of which are unnecessary: for example, this platform does not have separate LSE and HSE clocks, and the USB peripheral can only be driven from the PLL, not the HSI oscillator), as well as some compiler-specific macros that seem to assume the drivers are being compiled with ARM CC, which causes problems for gcc.

To try to fix this, I have combined my previous implementation with the implementation currently here to create a working version. I'm happy to address any particular changes, but essentially I have:

 - Replaced the custom CMSIS headers with the [OEM versions](https://github.com/OpenNuvoton/NUC123BSP/tree/master/Library/Device/Nuvoton/NUC123/Include)

- Moved the clock initialization functions from the non-standard CLKv1 module into the base platform drivers (hal_lld.c & hal_lld.h)

- Added board drivers for the NUTINY-SDK-NUC123-V2.0 dev board, which I am testing all of this on

- Other misc. janitorial work to make everything compile & play nice

I have other drivers to contribute to this platform once this is resolved.


alexclewontin:
The Jenkins build failure is not related to this PR (fails somewhere during the STM32 build), and Jenkins doesn't actually test this platform.

EDIT: I have added a NUMICRO Jenkins stage. This isn't showing up in the checks for this PR, but would be helpful down the road.

fpoussin:
Hi,
Could you rebase? I've fixed CI and merged another PR.
Thanks!

alexclewontin:
@fpoussin Rebased, and passing CI checks (though of course the Jenkins build doesn't mean anything yet)

fpoussin (approved):

fpoussin:
It's fine, we can fix CI after the merge

fishman:
this actually completely broke my board

fishman:
@alexclewontin I use the clock driver for my rgb matrix driver, which is no longer working now.

https://github.com/fishman/qmk_firmware/blob/duckyone2mini/keyboards/ducky/one2mini/mbi5042gp.c#L220

https://github.com/qmk/qmk_firmware/pull/11112/files?file-filters%5B%5D=.c#diff-ada687378e5dbbe691a8a9c5987ddacac87668044f33edc49107f6bef30bc53bR207

fishman:
Basically without providing a PWM driver interface disabling the functions breaks the ability to setup PWM signals

alexclewontin:
I can check out implementing a barebones PWM module later today, but fwiw, the CLK module is not a standard part of the ChibiOS HAL, and no other platform offers a CLK module. The most maintainable and extensible thing in the long term would probably be to write drivers using the standard structure

fishman:
@alexclewontin I've started to work on this, maybe you can point me in the right direction?

https://github.com/fishman/ChibiOS-Contrib/blob/numicro-pwm/os/hal/ports/NUMICRO/LLD/TIMv1/hal_pwm_lld.c

alexclewontin:
@fishman The problem with that driver is that it does not conform at all to the ChibiOS HAL API. I'm about to open a draft pull request with a (hopefully working) driver, which you can use downstream at least temporarily

fishman:
no i'm saying that i've started working on one that does the proper API. Would have been a good opportunity to understand the components here. But I guess if you're already done then it doesn't matter anymore.

alexclewontin:
@fishman apologies, I just went ahead and banged it out this afternoon so the QMK projects downstream (including yours) could get started using it. It just doesn't make sense to have two people duplicate efforts. But the driver is far from done: I've added a list of things that need to happen on it still in the draft PR, and I'm happy to discuss/change any design decisions I made in what I've posted so far.

fishman:
@alexclewontin looks good

alexclewontin:
@fishman please feel free to tackle anything on that list. I'll spend some more time on it tomorrow, but the things that are left are things that it would probably be better to have a second set of eyes on

PR #234, opened 2020-12-08
NUMICRO: add Ein Terakawa's i2c driver

fishman:
Add numicro i2c driver


fishman:
He'll push this himself

PR #236, opened 2020-12-09
[NUC123SD4AN0] Barebones PWM module for downstream compatibility

alexclewontin:
This is a barebones PWM module to allow downstream QMK projects to use the PWM peripheral. There is more work to do on it, but I wanted to at least open the PR as a draft to communicate with others who are also working on this module.

### DIFF
--- a/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.c
+++ b/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.c
@@ -16,7 +16,9 @@
 
 #include "hal.h"
 
+#if HAL_USE_PAL
 const PALConfig pal_default_config;
+#endif
 
 /**
  * @brief   Board-specific initialization code.
--- a/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.h
+++ b/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.h
@@ -27,7 +27,6 @@
 #define NUC123SD4AN0
 #define BOARD_NAME "NUTINY SDK NUC123 V2.0"
 
-
 #define NUC123_LSE_BYPASS TRUE
 /*
  * Board specific settings.
--- a/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.mk
+++ b/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.mk
@@ -4,9 +4,5 @@ BOARDSRC = $(CHIBIOS_CONTRIB)/os/hal/boards/NUTINY-SDK-NUC123-V2.0/board.c
 # Required include directories
 BOARDINC = $(CHIBIOS_CONTRIB)/os/hal/boards/NUTINY-SDK-NUC123-V2.0
 
-# Define linker script file here, as the linker script is per chip variant, which should 
-# remain constant per board
-LDSCRIPT = $(STARTUPLD_CONTRIB)/NUC123SD4AN0.ld
-
 ALLCSRC += $(BOARDSRC)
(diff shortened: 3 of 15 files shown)


alexclewontin:
@fishman Probably best to base your work on this instead of duplicating it. More to be done here:

- testhal test case
- Periodic callbacks. Since PWM0, PWM1, PWM2, and PWM3 each have their own down-counter, how do we trigger a periodic callback. Is this worth splitting them into 4 separate drivers, each with a single channel? Then how do we handle frequency?
- Pin muxing: there are multiple pins that can be muxed to each channel. How should the driver be told which one? Should it just assume the user will do it outside the driver code?
- Inversion: there are two separate bits that control inversion, and I have not yet figured out what they do, so that still needs addressing.
- Deadzones: probably the lowest priority

fishman:
If we split them into four drivers then frequency can be handled as a fraction of the base frequency no ? The main difference here between this and most other pwm drivers in chibios is that the other ones all have more than one interrupt handler and the nuc123 although has one right ؟

alexclewontin:
~~Yeah it probably makes the most sense to treat them as four separate driver objects. The ISR will have to differentiate between them in software as opposed to hardware, but that's not the end of the world. What really convinces me is this description from the [HAL docs](https://chibiforge.org/doc/20.3/hal/group___p_w_m.html#ga59a575307a1a1dceecbd15e70963e36d):~~

~~"This driver abstracts a generic PWM timer composed of:~~

- ~~A clock prescaler.~~
- ~~A main up counter.~~
- ~~A comparator register that resets the main counter to zero when the limit is reached. An optional callback can be generated when this happens.~~
- ~~An array of PWM_CHANNELS PWM channels, each channel has an output, a comparator and is able to invoke an optional callback when a comparator match with the main counter happens."~~

~~Based on that description, we pretty clearly have 4 single-channel PWM units, with some connected clock generation.~~

~~The frequency thing is probably right: this will put a little limitation on what frequencies are can work, as Timers 0 & 1 share PPR::CP01 and can only be differentiated by CSR::CSRn so they have to be within a factor of 16 of each other (and the same for 2 & 3).~~

I think I've changed my mind.

The basic pattern of the HAL is that each driver is a state machine (can be seen at above link):
```
UNINITIALIZED --peripheralInit()--> LOW POWER <--peripheralStop()/peripheralStart()--> READY
```

The problem here is that in the CLK controller (i.e. `CLKSEL1/2`, `APBCLK`) you can only configure the PWMs 2 at a time. Even worse, in `SYS->IPRSTC2`, you can only reset them all at once, not individually. Thus, I think to separate them into 4 driver objects, you'd need to have like, a file-static counting semaphore to keep track of how many drivers are currently enabled, and only do the clock starting & stopping when it transitions from 0 to 1 & vice versa. This seems like it goes against the design pattern of the HAL.

It probably makes more sense to treat the different timers as channels within one PWM driver. We could keep all the timers' periods in phase just by using the same clock source and starting & stopping them all together (they'd run constantly while the peripheral is in the READY state, and the individual channels would be enabled & disabled using the POE control), and artificially designate one of them as the "Periodic channel" which triggers the periodic callback. 

attn: @fishman

alexclewontin:
@fpoussin (or other maintainer): I think this module is ready to merge. The LGTM analysis is getting tripped up somehow, but I cannot find any meaningful error in the logs, and I don't think LGTM even tests the NUC123 platform currently, so I am a little confused as to what exactly is wrong. Please let me know if I need to fix something.

fpoussin (approved):

fpoussin:
Looks good, thanks!

PR #239, opened 2020-12-14
[NUC123] Serial driver

alexclewontin:
HAL driver module for serial communications on the NUC123, and a testhal case.


fpoussin (approved):

fpoussin:
Thank you

PR #240, opened 2020-12-14
[NUC123] USB Driver rewrite & testhal

alexclewontin:
As it was written, the USB driver in associated with the NUC123 was not actually working (when used with this testhal case at least). Here is another version which does in fact work with the test case.

Apologies for the many PRs, and thank you for your work in merging them!

### DIFF
--- a/os/hal/ports/NUMICRO/LLD/USBv1/driver.mk
+++ b/os/hal/ports/NUMICRO/LLD/USBv1/driver.mk
@@ -1,9 +1,11 @@
 ifeq ($(USE_SMART_BUILD),yes)
 ifneq ($(findstring HAL_USE_USB TRUE,$(HALCONF)),)
 PLATFORMSRC += $(CHIBIOS_CONTRIB)/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
+PLATFORMASM += $(CHIBIOS_CONTRIB)/os/hal/ports/NUMICRO/LLD/USBv1/usb_memcpy.S
 endif
 else
 PLATFORMSRC += $(CHIBIOS_CONTRIB)/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
+PLATFORMASM += $(CHIBIOS_CONTRIB)/os/hal/ports/NUMICRO/LLD/USBv1/usb_memcpy.S
 endif
 
 PLATFORMINC += $(CHIBIOS_CONTRIB)/os/hal/ports/NUMICRO/LLD/USBv1
--- a/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
+++ b/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
@@ -1,6 +1,5 @@
 /*
-    ChibiOS - Copyright (C) 2017 Frank Zschockelt
-    ChibiOS - Copyright (C) 2019 /u/KeepItUnder
+    Copyright (C) 2020 Alex Lewontin
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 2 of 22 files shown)


fpoussin (approved):

fpoussin:
Thanks!

fishman:
Unfortunately this too completely breaks USB for me :( also I don't know how downstream usually handles handles ASM sources, but  adding usb_memcpy doesn't work unless I add it to the startupasm

alexclewontin:
@fishman hmm, sounds like the downstream isn't respecting the  ChibiOS build system.

Can you link me to it? I can look at it and try to see exactly what is happening. It shouldn't be hard to implement, just adding a rule that correctly targets ASM files.

The ASM is present because the USB SRAM enforces access width constraints above and beyond the rest of the platform, and my reasoning was, if you need something done in ASM a certain way, you should probably just be writing it in ASM, instead of hoping all optimizing compilers handle it the way you want.

fishman:
I just fixed it https://github.com/qmk/qmk_firmware/pull/11219 . Yes they don't use the ChibiOS build. So the build is fixed but the keyboard still doesn't work anymore for me.

alexclewontin:
Good work, yeah that's exactly what needs to happen. In terms of the USB not working, that's obviously concerning. The testhal case works for me with this driver, whereas it did not with the old one (which wasn't even complete: several functions weren't implemented). My best guess is that it has something to do with config you are using to build. That's still a problem however, as the driver should either complain at compile time, or work. Can you point me to the repo/branch you are trying to build from?

EDIT: When you say doesn't work, do you mean even after #241?

alexclewontin:
@fishman Just as an update: I have built a version of qmk for the dev board, and the USB is indeed not working for some reason. I'm digging into it with the dev board ICE debugger and Wireshark, and what I can see so far is that for some reason the firmware is sending the wrong descriptor in response to requests, which is derailing the enumeration process. I'm gonna keep looking and with any luck (though no promises) will have a diagnosis/fix tonight or tomorrow

fishman:
@alexclewontin do you have a writeup anywhere of how you're debugging this? I'm pretty much doing trial and error here, which isn't the best way to work on this.

alexclewontin:
@fishman I don't have anything formal, although I'll consider putting something together. In the meantime, this isn't an exhaustive guide, but a few points:

- I have a couple of [these](https://www.digikey.com/catalog/en/partgroup/nuc123-evaluation-board-nutiny-sdk-nuc123/41132) which are super useful for debugging. It allows you to set breakpoints, step through code, and examine memory & peripheral state in real time. I interface with it using OpenOCD & gdb, but you can also use it with the Keil uVision IDE debugger & some Nuvoton plugins. (I have a couple for no reason other than they're pretty fragile). To generalize a little bit, when working on a new platform my first step would pretty much always be to get on-chip debugging working, whether that be via a dev board or a standalone adapter hooked up to my target. 
- For slower busses (UART, I2C, I2S, SPI, CAN, etc...) a protocol/logic analyzer is great. I don't have much comparative experience, so I will refrain from making any specific recommendations, but being able to see what is going on at a bus & transaction level simultaneously is very helpful. For looking at control flow, the logic analyzer function is also very helpful: you can set up GPIO pins as flags, and have them toggle at specific points in the program, and monitor their state.
- For faster busses like USB, looking at the bus level is, frankly, difficult. I know Sigrok will do protocol decoding from a logic analyzer signal, but given you need to sample at ~48 MHz for USB Full Speed, your capture window will likely be limited, unless you have a USB-specific protocol analyzer (which I do not). At the packet level, Wireshark is invaluable. How exactly you set up captures depends on your platform, but you should end up with a real-time view of the packets being sent and received, and a dissection of them into fields. 

In this particular case, I built a version of qmk targeted at the dev board (I wrote a dummy GPIO driver that never activated the peripheral just to avoid needlessly driving and reading floating pins). I plugged in the USB to my computer and looked at the packets in Wireshark which is how I figured out that the MCU was sending a device descriptor with an incorrect `bLength` Then I used on-chip debugger to step through to figure out where the bad data was coming from.

As another piece of specific advice, I cannot emphasize enough how important it is to truly understand the protocol that you are trying to debug. In this case, it was super important to understand USB enumeration (the host first requests 8 bytes of the device descriptor. This includes the total length. The host then rerequests for the total length. Then on to the configuration descriptor...) For digging into the specifics of USB at all levels, http://www.usbmadesimple.co.uk/ and https://www.beyondlogic.org/usbnutshell/usb1.shtml are both great resources. They overlap some but are also very complementary.

PR #241, opened 2020-12-15
NUC123 change USBEndpointConfig

fishman:
USBEndpointConfig should have the same list of arguments as every other USB driver to allow downstream apps to use the same initializer


alexclewontin:
Yep, fair enough. Annoyingly, those fields aren't included in the main repo's template files, and I think they are just for backwards compatibility, but we should try to honor that. Good call

fpoussin (approved):

fpoussin:
Looks good, thanks!

PR #242, opened 2020-12-16
NUC123 USB driver: bug fix

alexclewontin:
This PR addresses a problem @fishman reported in the updated USB driver: in situations where memory was being copied between the USB SRAM buffer and general memory, and one block was word aligned, and one block is not, there was an off-by-one error in the index used for copying.


fpoussin (approved):

fpoussin:
Thanks!

fishman:
@alexclewontin just rebuilt my keyboard firmware. unfortunately i still don't get any keypresses detected

alexclewontin:
@fishman does the host computer recognize that a keyboard is attached?

fishman:
@alexclewontin nope

PR #243, opened 2020-12-18
[HAL/LPC] Update USB HAL Driver to resolve an issue where sometimes it causes control EP to stall

Codetector1374:


fpoussin (approved):

fpoussin:
Thanks!

PR #246, opened 2020-12-29
Teensy 4.x support

stapelberg:
I’m creating this Pull Request for visibility, so that other contributors can see this work in progress and don’t waste time duplicating work.

This is not in any shape to be merged yet, but it works on real hardware: https://github.com/kinx-project/kint/issues/5#issuecomment-751999435


fpoussin:
any update?

stapelberg:
> any update?

Life got in the way :) This is still on my list, but I can’t promise when I’ll get around to cleaning these commits up. Hopefully within the next few weeks.

stapelberg:
Status update: started the cleanup effort. I pushed a few commits that switch back from the jammed-in teensy 4 startup code to the ChibiOS startup code, and got it to work on the Teensy 4.1.

Further cleanup still required before this is ready to review.

stapelberg:
Status update: I cleaned up the changes on the git commit level:
status

However, there still are a few things to clean up within these commits. I’ll report back when I get a chance.

stapelberg:
I think this is ready now!

I have replaced my custom changes with calls into the NXP MCUXpresso SDK and USB middleware where possible. This should be a much more robust base than a custom implementation.

I added a git submodule for https://github.com/NXPmicro/mcux-sdk, which will allow us to pull in new changes easily. The USB middleware was copied, but NXP tells me they plan to publish that code in a git repository, too: https://github.com/NXPmicro/mcux-sdk/issues/12

I tested my changes by building the TEENSY4_1 demo I added and running it on the Teensy 4.1 and MIMXRT1060-EVK eval board. I also successfully built the QMK firmware and ran it on both boards.

stapelberg:
@fpoussin did you have a chance to take a look yet? It’d be great to discuss this soon, as long as I have all the details in my head still :)

fpoussin:
Not yet, I'll have a look today.

fpoussin:
git modules or any folders in root are a no-go, they need to be in /ext
extra libs need to be in compressed 7z archives as well (see /ext)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "mcux-sdk"]
+	path = mcux-sdk
+	url = https://github.com/NXPmicro/mcux-sdk
```
fpoussin, line 3: This needs to go in /ext

fpoussin (changes requested):

.gitmodules
stapelberg: Done

stapelberg:
> extra libs need to be in compressed 7z archives as well (see /ext)

I see SEGGER-RTT_6.44i+SystemView_2.52h.7z but no other information.

Given that the files are required for building, wouldn’t it be inconvenient to have them only in a 7z archive? Or am I misunderstanding what you mean?

fpoussin:
> > extra libs need to be in compressed 7z archives as well (see /ext)
> 
> I see SEGGER-RTT_6.44i+SystemView_2.52h.7z but no other information.
> 
> Given that the files are required for building, wouldn’t it be inconvenient to have them only in a 7z archive? Or am I misunderstanding what you mean?

The rationale is that the majority of users won't be needing it, so in order to keep the repo as small/simple as possible it is compressed.
ChibiOS's HAL is not meant to be a wrapper around other HALs (you'd use them directly instead) so the structure is not organised to host this kind of big libraries.
If it's not too large it's fine to keep it in the current state (uncompressed)

stapelberg:
> The rationale is that the majority of users won't be needing it, so in order to keep the repo as small/simple as possible it is compressed.
> ChibiOS's HAL is not meant to be a wrapper around other HALs (you'd use them directly instead) so the structure is not organised to host this kind of big libraries.
> If it's not too large it's fine to keep it in the current state (uncompressed)

Understood.

I think the model of using Git submodules like for `ext/mcux-sdk` is useful here: most users won’t need to check out these submodules and can get a lean repository.

Unfortunately, NXP does not yet provide its USB middleware in a git repository, hence the need to copy the files. It’s only 70 files, though, so that should be fine as a temporary measure until NXP provides a git repository (which they said they would).

I’d prefer to leave them as-is instead of having them in a 7z file. I think leaving them as-is will make using them in QMK easier.

fpoussin (approved):

fpoussin:
Sounds good, can I merge this?

stapelberg:
> Sounds good, can I merge this?

Yes, this is good to merge :)

Thank you!

fpoussin:
Merged, thank you!

PR #257, opened 2021-02-04
USB MSD fix, caused crash on connecting to Windows or MacOS

pierstitus:
Reimplement msd_request_hook, probably this code was written for an old ChibiOS version?

Requests were incorrectly handled, causing crashes on connecting to Windows or MacOS.

### DIFF
--- a/os/hal/src/hal_usb_msd.c
+++ b/os/hal/src/hal_usb_msd.c
@@ -33,7 +33,7 @@
 /*===========================================================================*/
 
 #define MSD_REQ_RESET                   0xFF
-#define MSD_GET_MAX_LUN                 0xFE
+#define MSD_REQ_GET_MAX_LUN             0xFE
 
 #define MSD_CBW_SIGNATURE               0x43425355
 #define MSD_CSW_SIGNATURE               0x53425355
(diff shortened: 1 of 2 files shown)


fpoussin (approved):

fpoussin:
Looks good, thanks!

mck1117:
it looks like this actually introduced a bug - MSD_REQUEST_GET_MAX_LUN arrives with the DEV2HOST bit set, so the current code actually misses it, and stalls the request erroneously.

I have a fix in #258

mck1117:
Here's a usbpcap trace of the request being ignored (and windows retrying several times):
image

pierstitus:
Oops, my bad. Great you found out. 

Have you ever seen the Bulk-Only Mass Storage Reset in the wild? I couldn't find too much information about those, only that they're not used often. In my code I assume that no reset is necessary, and behave as if the reset is finished immediately. Adding a callback to user code that may just issue a system reset might make sense, but if it happens so rarely that it's hard to test it's probably not worth the effort.

mck1117:
I don't think I've ever actually seen one happen, but presumably it'd need something to go wrong for it to happen.  I messed around for a while with a breakpoint set on that line and never saw it happen.

pierstitus:
In the linux source I found two calls that issue the reset, but yeah, there should be something wrong for it to happen.

Those files both have a call to transport_reset, which points to the Bulk-Only Mass Storage Reset function usb_stor_Bulk_reset
https://github.com/torvalds/linux/blob/3db1a3fa98808aa90f95ec3e0fa2fc7abf28f5c9/drivers/usb/storage/transport.c
https://github.com/torvalds/linux/blob/ab91292cb3e9f43d9c6839d7572d17b35bc21710/drivers/usb/storage/scsiglue.c

I guess for most cases the current solution is fine, and if not there's at least some comments in the code.

PR #262, opened 2021-02-14
[NUC123] EFL subsystem driver, and updated flash control

alexclewontin:
This commit includes a driver for the NUC123's embedded flash programmer, and corresponding testhal case. It also updates the platform linker script to better reflect the layout of the various blocks of embedded flash. Finally, it adds variable representations of the persistent configuration registers, which can be programmed when flashing the .elf file. The linker script places these accordingly in memory.


testhal/NUMICRO/NUC123/NUTINY-SDK-NUC123-V2.0/EFL/shcfg.c
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  uint8_t buf[128];
+  size_t  n = 128;
+  mfs_error_t err = mfsReadRecord(&mfsd, rid, &n, buf);
+  switch (err) {
+    case MFS_WARN_GC:
+    case MFS_WARN_REPAIR:
+    case MFS_NO_ERROR:
+      chprintf(chp, "%s\n", buf);
```
elfmimi: buf must be zero terminated. i.e. buf[n] = 0;
and it will be nice to increase sizeof(buf) by 1 or decrease n by 1.

elfmimi:
I want these lines to be activated only with certain conditions.
```C
volatile const uint32_t config0 __attribute__((used, unused, section(".nuc123_config0"))) = NUC123_CONFIG0;
volatile const uint32_t config1 __attribute__((used, unused, section(".nuc123_config1"))) = NUC123_CONFIG1;
```

Yes, it may be useful for testing purposes. But usually you want to avoid accidental overwriting of CONFIG registers.
In addition, if config sections exist in the elf files conversion to simple binaries by objcopy will get harder, while it is still doable.

alexclewontin:
> I want these lines to be activated only with certain conditions.
> 
> ```c
> volatile const uint32_t config0 __attribute__((used, unused, section(".nuc123_config0"))) = NUC123_CONFIG0;
> volatile const uint32_t config1 __attribute__((used, unused, section(".nuc123_config1"))) = NUC123_CONFIG1;
> ```
> 
> Yes, it may be useful for testing purposes. But usually you want to avoid accidental overwriting of CONFIG registers.
> In addition, if config sections exist in the elf files conversion to simple binaries by objcopy will get harder, while it is still doable.

I've added a switch to enable them, that defaults to FALSE. This also involved adding a couple of `#error` messages, and changing the default data flash values to reflect the default erased CONFIG values.

The EFL driver's behavior depend on their value, whether they're explicitly set or not. When you don't set them, you still make a decision about their value, it just is implicit as opposed to explicit, and if you have set it through other means, trying to use the EFL driver with data flash could fail in a non-obvious way. To solve this, I've required that the CONFIG registers be set by the HAL if the EFL driver is going to access the data flash.

testhal/NUMICRO/NUC123/NUTINY-SDK-NUC123-V2.0/EFL/shcfg.c
alexclewontin: I just added a max width to the format string, should solve the problem. As long as there is no overflow, the null termination comes from when it was stored.

elfmimi:
I'm sorry to mention this.
I consider that NUC123_CONFIG_ENABLED == FALSE does not imply CONFIGs are at their default 0xFFFFFFFF state.
I consider NUC123_CONFIG_ENABLED == FALSE as "trust the configuration given by that human being" .
so these 3 #error-s look somewhat like overstepping to me. https://github.com/ChibiOS/ChibiOS-Contrib/pull/262/commits/c0d0b5bac2e94a0b7f0633cbaa009a4c9c9e815a
could you loosen those restrictions?

alexclewontin:
> I'm sorry to mention this.
> 
> I consider that NUC123_CONFIG_ENABLED == FALSE does not imply CONFIGs are at their default 0xFFFFFFFF state.
> 
> I consider NUC123_CONFIG_ENABLED == FALSE as "trust the configuration given by that human being" .
> 
> so these 3 #error-s look somewhat like overstepping to me. https://github.com/ChibiOS/ChibiOS-Contrib/pull/262/commits/c0d0b5bac2e94a0b7f0633cbaa009a4c9c9e815a
> 
> could you loosen those restrictions?

I disagree. The driver either gets permission to manage the CONFIG registers or not, but if the user wants to manually manage the CONFIG registers, the driver should not pretend that it knows the values. This kind of misalignment between software and hardware state is the most insidious kind of problem to debug.

For example, if a user sets `NUC123_DATAFLASH_SIZE` to a custom value, but does not set `NUC123_CONFIG_ENABLED` to true, they might reasonably expect that the data flash size gets changed. When it doesn't, this will lead to silent errors in the EFL driver, and I think avoiding silent errors is one of the top priorities.

EDIT: I can look at changing it so that it does not even assume the 0xFFFFFFFF, but I think that involves tightening the restrictions, not loosening them. For instance, the size of the APROM requires knowing the size of the data flash. So removing the default assumption about the default 4096 bytes of data flash means requiring that CONFIG be enabled to use the EFL driver for APROM as well.

fpoussin:
Hi,
Let me know when it's ready for review,
Thanks!

alexclewontin:
@fpoussin will do, I have one more piece coming after the discussion.

@elfmimi on further consideration, I think we have the same concerns but are approaching it differently. I'm working on a slight modification that involves dynamically reading the CONFIG value in certain cases, so that it does not need to be set at compile time, but the EFL driver still has the accurate values. An added bonus here is that if the CONFIG values are reprogrammed using the EFL driver, it doesn't mess things up. I'll push it here soon.

elfmimi:
just in case, I leave a note here. value of CONFIG1 can be obtained by FMC->DFBADR .
if DFVSEN = 1 then FMC->DFBADR is 0x1F000 .
bad thing is , when DFVSEN = 0 and DFEN = 1  ~(0)~  we would desire FMC->DFBADR to be 0x11000 or 0x9000 depending on flash size, 
but it does not behave that way. FMC->DFBADR just shows the value of CONFIG1 .
that means we still need to access CONFIG0 through ISP or FMC.

alexclewontin:
@elfmimi Take a look at the latest commit. I think it addresses your concerns (CONFIG register values off by default) while preserving the function of the EFL driver

elfmimi:
> I think we have the same concerns but are approaching it differently.

This is very true.

It's not like I'm trying to delay the merge of this PR.
This PR is about adding EFL subsystem, which is to add support for FMC, Flash Memory Controller, in Nuvoton's term.
The problem is that a little portion of this PR which deals with CONFIG is interfering with plain vanilla workflow developers are accustomed to.
If you really truly have to include consistency check into this RP anyhow, maybe it could be achieved by letting the actual value of CONFIG0 and CONFIG1 defined as symbols in mcuconfig.h . and that values must go into the elf file AS IS, if required.
My suggestion is not changed from before. Let's refrain from making a big deal out of this. at least at this point.

Second thing, dynamic acquisition of configuration. This is one style of usage intended by Nuvton's hardware design.
Despite your hard work, NUC123_EFL_DYNAMICALLY_CHECK_CONFIG is still not fully functional because it is not deducing dataflush_size property. note: there is a way to detect flash memory size using FMC_ISPCMD_READ .
Again, I suggest backing down a bit and adding this later when it gets ready.

testhal/NUMICRO/NUC123/NUTINY-SDK-NUC123-V2.0/EFL/shcfg.c
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  uint8_t buf[128];
+  size_t  n = 128;
+  mfs_error_t err = mfsReadRecord(&mfsd, rid, &n, buf);
+  switch (err) {
+    case MFS_WARN_GC:
+    case MFS_WARN_REPAIR:
+    case MFS_NO_ERROR:
+      chprintf(chp, "%.128s\n", buf);
```
elfmimi: It is still erroneous. Try putting string with length of multiple of 4, and it actually behaves wrong when getting.
I was expecting this.
```c
chprintf(chp, "%.*s\n", n, buf);
```

alexclewontin: Fixed

alexclewontin:
> The problem is that a little portion of this PR which deals with CONFIG is interfering with plain vanilla workflow developers are accustomed to.
> If you really truly have to include consistency check into this RP anyhow, maybe it could be achieved by letting the actual value of CONFIG0 and CONFIG1 defined as symbols in mcuconfig.h . and that values must go into the elf file AS IS, if required.
> My suggestion is not changed from before. Let's refrain from making a big deal out of this. at least at this point.
> 
> Second thing, dynamic acquisition of configuration. This is one style of usage intended by Nuvton's hardware design.
> Despite your hard work, NUC123_EFL_DYNAMICALLY_CHECK_CONFIG is still not fully functional because it is not deducing dataflush_size property. note: there is a way to detect flash memory size using FMC_ISPCMD_READ .
> Again, I suggest backing down a bit and adding this later when it gets read

@elfmimi I've added the missing bit so that dataflash_size is computed (at line 255 in hal_efl_lld.c). 

Other than that, I do not really understand what is interfering with the workflow. At this point, the CONFIG0 and CONFIG1 values are not included unless they are specifically enabled. There is nothing that requires developers to enable them, and unless they specifically seek them out, they will not know they are there. I do not see the disadvantage of including them and disabling them by default, but there are advantages to giving developers the option to enable however (improving performance and reducing code size)

testhal/NUMICRO/NUC123/NUTINY-SDK-NUC123-V2.0/EFL/cfg/mcuconf.h
```diff
@@ -0,0 +1,42 @@
...
+ * HAL driver system settings.
+ */
+#define NUC123_PLL_ENABLED TRUE
+#define NUC123_PLLSRC NUC123_PLLSRC_HSE
+
+#define NUC123_SERIAL_CLKSRC NUC123_SERIAL_CLKSRC_HSI
+
+#define NUC123_CONFIG_ENABLED TRUE
+#define NUC123_DATAFLASH_ENABLED TRUE
+#define NUC123_DATAFLASH_SIZE 4096
```
elfmimi, line 34:
```suggestion
#define NUC123_CONFIG_DATAFLASH_ENABLED TRUE
#define NUC123_CONFIG_DATAFLASH_SIZE 4096
```

os/hal/ports/NUMICRO/LLD/FLASHv1/hal_efl_lld.c
```diff
@@ -0,0 +1,690 @@
...
+
+#define NUC123_LDROM_SIZE 0x1000UL
+
+#define NUC123_EFL_CMD_ERASE     0x22UL
+#define NUC123_EFL_CMD_PROG      0x21UL
+#define NUC123_EFL_CMD_READ      0UL
+#define NUC123_EFL_CMD_CHIPERASE 0x26UL /* Undocumented */
+
+#if ((NUC123_CONFIG_ENABLED == FALSE) || (NUC123_EFL_ACCESS_CONFIG == TRUE)) && \
+    (NUC123_EFL_ACCESS_APROM == TRUE) && (NUC123_EFL_ACCESS_DATAFLASH == TRUE)
```
elfmimi, line 52:
```suggestion
    ((NUC123_EFL_ACCESS_APROM == TRUE) || (NUC123_EFL_ACCESS_DATAFLASH == TRUE))
```

os/hal/ports/NUMICRO/LLD/FLASHv1/hal_efl_lld.c
```diff
@@ -0,0 +1,690 @@
...
+#define NUC123_EFL_DYNAMICALLY_CHECK_CONFIG TRUE
+#endif
```
elfmimi, line 54:
```suggestion
#else
#define NUC123_EFL_DYNAMICALLY_CHECK_CONFIG FALSE
#endif
```

os/hal/ports/NUMICRO/LLD/FLASHv1/hal_efl_lld.c
```diff
@@ -0,0 +1,690 @@
...
+    /* Enables the peripheral.*/
+    CLK->APBCLK |= CLK_AHBCLK_ISP_EN_Msk;
+
+#if NUC123_HCLK < NUC123_LFOM_UPPERBOUND
+    FMC->FATCON |= FMC_FATCON_LFOM_Msk;
+#elif NUC123_HCLK < NUC123_MFOM_UPPERBOUND
+    FMC->FATCON |= FMC_FATCON_MFOM_Msk;
+#endif
+
+#if (NUC123_EFL_ACCESS_APROM == TRUE) || (NUC123_EFL_ACCESS_DATAFLASH == TRUE)
```
elfmimi, line 169:
```suggestion
#if (NUC123_EFL_ACCESS_APROM == TRUE)
```

os/hal/ports/NUMICRO/LLD/FLASHv1/hal_efl_lld.c
```diff
@@ -0,0 +1,690 @@
...
+{
+  uint32_t ispcon;
+  if (eflp->state == FLASH_READY) {
+    SystemUnlockReg();
+    ispcon = FMC->ISPCON;
+
+    ispcon &= ~FMC_ISPCON_ISPEN_Msk;
+      /* Disables the peripheral.*/
+
+#if (NUC123_EFL_ACCESS_APROM == TRUE) || (NUC123_EFL_ACCESS_DATAFLASH == TRUE)
```
elfmimi, line 203:
```suggestion
#if (NUC123_EFL_ACCESS_APROM == TRUE)
```

os/hal/ports/NUMICRO/LLD/FLASHv1/hal_efl_lld.c
```diff
@@ -0,0 +1,690 @@
...
+      dataflash_size = NUC123_FLASH_SIZE - (uint32_t)dfbaddr;
+    }
+  }
+
+  FMC->ISPCON = ispcon;
+
+#else
+
+  dataflash_size = NUC123_CONFIG_DATAFLASH_SIZE;
+  dfbaddr = NUC123_DFBADDR;
```
elfmimi, line 264:
```suggestion
  dfbaddr = (void *)NUC123_DFBADDR;
```

elfmimi (approved):
Great! Now it works fine. At least I can live with this.

elfmimi (approved):

fpoussin:
@elfmimi feel free to merge it

elfmimi:
@fpoussin I'd like to. But I'm still seeing this.
MergeBlocked

fpoussin:
It should be ok now

elfmimi:
gonna click "Update branch" and try to proceed.
BranchOutOfDate

elfmimi:
Thanks!

PR #263, opened 2021-02-24
don't deref null pointer on MSD LUN request

mck1117:
Calling `usbSetupTransfer(usbp, 0, 1, NULL);` will deference a null pointer.  On an STM32 this silently actually works because `0x00000000` is a valid address to read - it reads the very bottom of flash or RAM, but it might not contain the `0` we want to transmit!


fpoussin (approved):

fpoussin:
Thanks :)

PR #265, opened 2021-03-05
Allow configuration of MSD thread

mck1117:
- Set MSD thread priority
- Set MSD thread stack size.  Some block devices (this one: https://github.com/rusefi/rusefi/pull/2420/files#diff-03042e564d3ce600e21979e4123bd93657ffdb926554e51e0deed52140518b94) might need more memory than the default 256 bytes.


fpoussin (approved):

fpoussin:
Thanks!

PR #270, opened 2021-04-03
Follow-up fix to alexclewontin's nuc123-kvs-example

elfmimi:
The re-works of change-requests which were left behind in https://github.com/ChibiOS/ChibiOS-Contrib/pull/262 .


fpoussin (approved):

elfmimi:
Thanks!

PR #272, opened 2021-04-05
Fix MIMXRT1062/LLD/USBHSv1: increase USB_DEVICE_CONFIG_ENDPOINTS from 4 to 7 (== USB_MAX_ENDPOINTS)

stapelberg:
This is required to enable the debug console in QMK, which makes the device start using more than 4 endpoints.


fpoussin (approved):

fpoussin:
Thanks!

PR #266, opened 2021-03-16
[RISC-V] ECLIC Port and GD32VF103 HAL Support

KarlK90:
This pull request adds ChibiOS support for the RISC-V ECLIC architecture by Nucleisys as well as a HAL Port for the GD32VF103 MCU family which is based on this architecture.

The GD32VF103 has been developed by GigaDevice and is a largely API compatible STM32F103/F105/F107 "clone" with a RISC-V RV32IMAC core. 

(Some) Differences:
Clock tree is mix of STM32F103 and STM32F105/STM32F107
USB OTG instead of USB Device peripheral
I2C supports Fast Mode Plus
SPI supports TI mode and NSS pulse mode
USB Clock prescalers with values 2 and 2.5 that allow a core Sysclock of 96MHz and 120MHz (Out of spec but works fine)
USB-DFU Bootloader!

My approach was to copy the existing Chibios STM32F103 HAL drivers and rename every register to the correct GigaDevice definition. (I talked to Giovanni about this and this was the preferred approach https://forum.chibios.org/viewtopic.php?f=8&t=5758)

So far I have tested:

GPIO
PWM
UART
DMA
USB OTG
I2C

So far untested:

SPI
RTC
ADC
FLASH
GPT
ICU
CAN
WGD
DAC

The I2C peripheral driver has a timing bug and randomly locks-up that I will address in a different pull request, all other tested drivers work flawless (at least for me).

I also added board definitions for the sipeed longan nano board as well as a usb cdc demo ported from STM.

Please let me know If this is good to merge. :-)

PR #273, opened 2021-04-14
[GD32] Fix I2C Hangups

KarlK90:
As mentioned in #266 there where random hangups using the I2C peripheral. Moving the DMA enable sequence back into the init code and re-adding the spurious bus error detection code from stmf4 error fixed these. I didn't verified what change fixed the issue  but my guess would be that enabling the DMA request from an IRQ context and just before starting a transfer isn't the best idea.


fpoussin (approved):

fpoussin:
Thanks!

PR #274, opened 2021-04-16
[RISC-V ECLIC] Add R/W memory and instruction barrier after mstatus access

KarlK90:
Fast subsequent reads and writes to the mstatus csr lead to illegal instruction exceptions on the nucleisys bumblee core of the gd32vf103. This behavior only occurred in high load situations e.g. interrupt frequency of 5khz but reliably let to these errors.  Adding the fence and fence.i memory/instruction barriers solved the problem. There is some negligible performance impact.


fpoussin (approved):

fpoussin:
LGTM, thanks!

KarlK90:
Thank you for your quick review, it just turned out that there is still a problem. I'll do a follow up pr.

PR #275, opened 2021-04-17
[RISC-V ECLIC] Illegal instruction exception on MSTATUS access

KarlK90:
The first attempt to solve illegal instruction exceptions was made in #274. It seemed as this "fixed" the issue, but merely added delays in the code which prevented the error to appear in lucky circumstances. Interesting that this code worked in the first place.
    
Root cause for the expection where write attempts to `mstatus` in user privilege mode which raised the illegal instruction exception which is in spec with the risc-v privileged isa and documented by [nucleisys](https://doc.nucleisys.com/nuclei_spec/isa/core_csr.html#accessibility-of-csrs-in-the-nuclei-processor-core). The solution is to never enter user mode by forcing `mcause.mpp` to 0x3 before calling `mret` when exiting the interrupt handler for context switching.

### DIFF
--- a/os/common/ports/RISCV-ECLIC/chcore.h
+++ b/os/common/ports/RISCV-ECLIC/chcore.h
@@ -377,9 +377,7 @@ static inline void port_init(void) {}
  * @return              The interrupts status.
  */
 static inline syssts_t port_get_irq_status(void) {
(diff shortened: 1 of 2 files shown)


fpoussin (approved):

fpoussin:
Thank you!

PR #276, opened 2021-04-20
[GD32VF103] Add CRC driver

KarlK90:
Ported the STM32 CRC driver to GD32VF103. The chip has a fixed CRC unit like the STM32F103. One LED define was false in the Longan Nano Board that I fixed in this pr as well.


fpoussin (approved):

fpoussin:
LGTM, thank you!

PR #277, opened 2021-04-25
[RISC-V ECLIC] Fix t0 restore when exiting interrupt

KarlK90:
An oversight when arranging the code according to the nucleisys docs, t0 was overriden with the value of msubm and never actually restored. To fix the issue we restore the csrs after the general purpose registers. The offical docs [want it the other way around](https://doc.nucleisys.com/nuclei_spec/isa/interrupt.html#interrupt-process-modes) but this should be fine as well, as the interrupts are still globally disabled at this point.


fpoussin (approved):

fpoussin:
Thank you

PR #278, opened 2021-04-27
Teensy 4.1 demo: start USB serial shell after control line state

stapelberg:
Without this change, the ChibiOS shell prompt does not show up reliably on
Linux, and possibly other operating systems.

For a detailed walk-through covering the individual problems and causes, see:
https://michael.stapelberg.ch/posts/2021-04-27-linux-usb-virtual-serial-cdc-acm/

I verified that this approach works with:

1. Linux 5.10
2. Windows 10
3. macOS Big Sur on a MacBook Air M1

### DIFF
--- a/demos/MIMXRT1062/RT-TEENSY4_1/main.c
+++ b/demos/MIMXRT1062/RT-TEENSY4_1/main.c
@@ -15,6 +15,7 @@
 */
 
 #include "ch.h"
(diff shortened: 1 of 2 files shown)


fpoussin (approved):

fpoussin:
Thank you :)

PR #279, opened 2021-05-12
[RISC-V ECLIC] Add RV32E support and fix context switching in case of pre-empted interrupts

KarlK90:
This adds support for the RV32E ABI which is a stripped down version on the RV32I ABI with only 16 (x0-x15) registers instead of the full set of 32 (x0-x31) registers.

Even more important is that the port had a one very critical bug related to pre-empted interrupts and context switches, leading to corrupted run time behavior in rather rare circumstances. This behavior was triggered when a interrupt was taken by the core and another higher privileged interrupt would per-empt the current processed interrupt, like seen in the [docs of nucleisys](https://doc.nucleisys.com/nuclei_spec/isa/interrupt.html#clic-mode-interrupt-preemption). Everything was working fine in case no re-scheduling was necessary, as this would jump to  `_port_exit_from_isr` after handling the higher priority interrupt. This would restore the preempted interrupt and continue servicing it. But in case a re-schedule was necessary, leaving the higher priority interrupt would result in a context switch by jumping to `_port_switch_from_isr`. This would not fully service the lower priority interrupt in the first place and corrupt the stackpointer.

The solution is to fully service any interrupt that was taken before doing a context switch (aka. only switch at the tail of a interrupt chain). On RISC-V this is notoriously difficult as there is normally no way to tell if you are at the tail of a (possibly pre-empted) interrupt chain. Fortunately nucleisys implemented the non-standard [msubm register with the msubm.ptyp](https://doc.nucleisys.com/nuclei_spec/isa/core_csr.html#msubm) field that holds the processing mode before the current interrupt was taken, 0x0 stands for normal processing mode. Because msubm is stored and restored for every interrupt we take, we can determine if we are at the tail. If the current msubm.ptyp value is 0 we know that we didn't interrupt any other interrupt before, so we are permitted to do a context switch. Likewise if the current msubm.ptyp value is different from 0 we know that we where not in normal machine mode, so we must have interrupted a lower priority interrupt. Context switching is forbidden at this point.

After this fix I haven't encountered any more glitches or hang-ups. My goal it to write one of the devs at nucleisys to have a look if they find anything odd with this port code.

### DIFF
--- a/os/common/ports/RISCV-ECLIC/chcore.h
+++ b/os/common/ports/RISCV-ECLIC/chcore.h
@@ -183,15 +183,17 @@ struct port_extctx {
   uint32_t a3;
   uint32_t a4;
   uint32_t a5;
+  uint32_t mepc;
+  uint32_t mcause;
+  uint32_t msubm;
+#if !defined(__riscv_32e)
   uint32_t a6;
   uint32_t a7;
   uint32_t t3;
   uint32_t t4;
   uint32_t t5;
   uint32_t t6;
-  uint32_t mepc;
-  uint32_t mcause;
-  uint32_t msubm;
+#endif
 };
 
 /**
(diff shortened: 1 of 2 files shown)


fpoussin (approved):

fpoussin:
Looks good, thank you!

PR #283, opened 2021-07-03
[GD32VF103] Rename missed AFIO registers and free B4 pin on Longan Nano

KarlK90:
I forgot to rename the AFIO registers in the struct, this PR corrects that. 

Also the PIN B4 has the JTAG NJRST signal with a pull up enabled by default. On the longan Nano this PIN is broken out as a regular pin. So we just disable NJRST by default and make it available as a regular GPIO.


fpoussin (approved):

fpoussin:
Looks good, thanks

PR #286, opened 2021-07-12
Replace packed struct definition with one from CMSIS

josesimoes:


os/hal/include/usbh/defs.h
```diff
@@ -26,11 +26,9 @@
 
 #ifdef __IAR_SYSTEMS_ICC__
 #define PACKED_STRUCT PACKED_VAR struct
```
fpoussin, line 28: Is this still used?

fpoussin (dismissed):

fpoussin (approved):

PR #287, opened 2021-07-19
Don't include SW CRC if not requested

snazarkin:
I've got build failure when enable CRC module on STM32. The reason is SW CRC header is included by default  but not found by compiler if ChibiOS-Contrib/os/various/ is not included into search path.
The fix includes only requested LL header.


fpoussin (approved):

fpoussin:
Thanks

PR #288, opened 2021-07-23
[GD32VF103] Fix ADC0 DMA stream mapping

KarlK90:
GigaDevice numbering scheme starts from 0 so the correct peripheral is dma0 with channel 0. This apparently slipped through when porting.


fpoussin (approved):

fpoussin:
Thank you!

PR #289, opened 2021-08-01
NUMICRO: Workaround for NUC123 USB Endpoint 6 problem

elfmimi:
See ER_6000_NUC123AN_EN_Rev.1.04.pdf for detail of the errata.

Currently the work-around is off by default.

Add following line to mcuconf.h to enable the work-around.
`#define NUC123_USB_WORKAROUND TRUE`

Enabling the work-around will increases the number of bi-directional endpoints from 2 to 3 .

PR #290, opened 2021-08-02
NUMICRO: Fix trivial but hard to find errors in USB LLD

elfmimi:
Fix:
- possible buffer overrun in usb_serve_out_endpoint()
- corner case src-dst confusion in usb_memcpy()

### DIFF
--- a/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
+++ b/os/hal/ports/NUMICRO/LLD/USBv1/hal_usb_lld.c
@@ -210,7 +210,7 @@ static void usb_serve_out_endpoint(uint32_t epn)
 {
   USBDriver *const usbp = &USBD1;
   uint32_t mxpld = HW_OUT_EP(epn)->MXPLD;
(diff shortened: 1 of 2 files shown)


elfmimi:
Plz merge this one prior to https://github.com/ChibiOS/ChibiOS-Contrib/pull/289 .

alexclewontin:
~~I'm not sure I understand these changes. For example, the changes to usb_memcpy seem backwards to me. Can you explain why this is necessary?~~ 

~~Ah I see. There is definitely a mismatch between the order of the src and dest param between usb_memcpy.S and hal_lld_usb.c. However, I think if you want to correct this in usb_memcpy.S, you need to change it in multiple places. Maybe better to change the invocations in hal_lld_usb.c to match the existing order in usb_memcpy.S~~

Ah, wrong again. Sorry about that, it's late for me. The change to usb_memcpy.S looks correct. I'll look at it again tomorrow with fresh eyes, and maybe also suggest a couple of changes to the comments in the file.

alexclewontin (approved):
Changed a couple of erroneous comments in usb_memcpy.S. LGTM

elfmimi:
Nice. My eyes were not seeing comments.

PR #295, opened 2021-09-06
[GD32VF103] Correct linker scripts

KarlK90:
The startup code that initializes the mcu is moved to be the first entry in the vectors section, so it will always be the first to run after a reset.

Because of a typo (vectors instead of .vectors) the vector table would have been there, but was relocated into the text section. So this setup compiled by sheer luck into a working executable that had the `_start` function at the flash base `0x8000000` up to this point.


fpoussin (approved):

fpoussin:
Thank you!

PR #297, opened 2021-09-11
K60x improvements

fauxpark:
I don't have hardware on hand to test, but these changes allow for successful compilation of a simple QMK firmware for Teensy 3.5 (MK64FX512).

Mainly fixing some incorrect defines that seem to have assumed non-OTG USB, but the K60 family appears to be OTG capable.


fpoussin (approved):

fpoussin:
Thanks!

PR #298, opened 2021-09-13
RP2040: Add I2C driver

hanya:
Interrupt based I2C driver.

It is really hard to implement DMA based I2C driver for sequence of 8-bit data.
Because IC_DATA_CMD register does not trigger if DMA writes 8-bit data into it.
We need 16-bit width transfer by DMA to make the register work to transmit new data.
I do not want to prepare sequence of 16-bit width data for transmitting purpose.
If we do expand data from 8-bit to 16-bit to write by DMA, we need 2ch of DMA with interrupt.

But the problem is not only the data but other flags in IC_DATA_CMD have to be set related to location of data.
CMD and STOP bits of IC_DATA_CMD must be set to start new write or read process.
Its hard to set such flags by DMA, we need interrupt  to set them.
Therefore I would not implement DMA based I2C driver myself.


fpoussin (dismissed):

fpoussin (approved):

fpoussin:
Thanks!

PR #299, opened 2021-09-14
RP2040: Add linker script for FLASH with boot2

hanya:
Pico sdk has boot2 section to load vector at XIP + 256. Therefore vector section have to be placed at the specific location.
The linker script from ChibiOS specifies 1024 alignment in rules_code.ld script. We can not use this with boot2 loader.

The boot2 can be linked as follows in Makefile:
```
ULIBS = $(PICOSDKROOT)/build/src/rp2_common/boot_stage2/bs2_default_padded_checksummed.S
```

RP2040 does not have any internal flash. RPi pico has external 2M flash for code. But other variant has different size of flash.
If we can change length of the XIP section, flash1. For this purpose, FLASH_LEN variable is defined which can be specified in user's custom linker script.

```
/* custom.ld */
/* Specifies flash size. */
"FLASH_LEN" = 16M;

INCLUDE RP2040_FLASH.ld
```

In Makefile: 
```
STARTUPLD_CONTRIB = $(CHIBIOS_CONTRIB)/os/common/startup/ARMCMx/compilers/GCC/ld

LDSCRIPT = ./custom.ld

ULIBS += -L$(STARTUPLD_CONTRIB)
```

### DIFF
--- a/os/common/startup/ARMCMx/compilers/GCC/ld/RP2040_FLASH.ld
+++ b/os/common/startup/ARMCMx/compilers/GCC/ld/RP2040_FLASH.ld
@@ -1,0 +1,117 @@
+/*
+    ChibiOS - Copyright (C) 2006..2021 Giovanni Di Sirio
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


fpoussin (approved):

fpoussin:
looks good!

PR #306, opened 2021-11-13
[WIP] Further K64F modifications, attempting to get USB working

fauxpark:
Since my last PR #297 I now have a Teensy 3.5. Unsurprisingly the first set of changes didn't work, so it seems there are more tweaks required to get USB functioning.

For the most part I've been referencing the PJRC Teensyduino code as well as the existing Kinetis implementations already in the repo (particularly the K66), as they're all very similar.
A simple QMK firmware is able to run and respond to key presses (https://github.com/qmk/qmk_firmware/pull/14420) - I can reset it back to bootloader mode - however the K64F's USB peripheral refuses to cooperate. Looking at the data lines in a logic analyser, both of them appear to be pulled up for some reason, with no other activity.

@stapelberg @walkerstop since you've done some work getting the 3.6 running, would either of you have any insight into this? I must be missing something obvious...


stapelberg:
> @stapelberg @walkerstop since you've done some work getting the 3.6 running, would either of you have any insight into this? I must be missing something obvious...

Hm, no idea off the top of my head 🤔

Have you compared behavior with a known-working USB implementation on the Teensy 3.5? I don’t have a much better strategy than to compare behavior and any differences step by step…

fauxpark:
My expectation was that Teensyduino would be a good reference - it doesn't seem to have a whole lot of chip-dependent code except for the obvious things like register definitions, which I've already checked. So I figured the existing USB LLD would also be more or less correct given that it looks like it's adapted from the Teensyduino stuff, and is used not just for the K66, but the K20xx chips as well.

os/hal/ports/KINETIS/LLD/USBHSv1/hal_usb_lld.c
```diff
@@ -427,6 +427,12 @@ void usb_lld_init(void) {
 #error USB clock setting not implemented for this KINETIS_MCG_MODE
 #endif /* KINETIS_MCG_MODE == ... */
 
+#if defined(K64F)
+  /* Switch from default MCGPLLCLK to IRC48M for USB */
+  //SIM->CLKDIV2 = SIM_CLKDIV2_USBDIV(0);
+  //SIM->SOPT2 |= SIM_SOPT2_PLLFLLSEL_IRC48M;
```
fauxpark, line 433: So this is strange... I managed to get USB working by commenting out these lines, dropping the PLL clock frequency from 120MHz down to 96MHz, and setting `SIM_CLKDIV2_USBDIV` to divide it by 2. The weird part is that I found the following sentence in [AN4905](https://www.nxp.com/docs/en/application-note/AN4905.pdf#d51e5a1310_d11e168):

> The K64, K63, and K24 family devices **with 1 MByte of flash** were the first Kinetis MCUs to implement the crystal-less USB feature

(emphasis mine)

This is not mentioned anywhere else, and if it means what I think it means, then the MK64FX512 cannot use the IRC48M for crystal-less USB, but the MK64FX1M0 *can*.

fauxpark: I'm going to have to leave this for someone smarter than me... as of now, USB works, but I have no idea if it's the best possible configuration.

fpoussin:
Is this ready to be merged?
can you rebase it? tests were broken.

fauxpark:
Thanks, rebased.

fpoussin (approved):

fpoussin:
Thanks!

PR #308, opened 2021-11-30
Modified i2c peripheral configuration

itarze:
Update modified to chibios-21.6.x.


itarze:
I don't understand why the buildRP failed.

PR #315, opened 2022-02-15
[RP2040] i2c: Reduce intterupt frequency

hanya:
Use I2C FIFO for TX much more to reduce intterupt frequecy for FIFO empty request.
RX must be write CMD bit for every byte, so we can not do like TX for RX. Only applied to TX.


fpoussin (approved):

fpoussin:
Thanks!

PR #316, opened 2022-03-02
[RP2040] Init USB out endpoints on transfer start

KarlK90:
@hanya I found an error in the USB driver, when starting new out transfers the out endpoint has to be configured immediately or otherwise a data sequence error is generated. The actual fix is on line 873 in `usb_lld_start_out()` Could you confirm / review my changes? 

The only notable other change was the condition for detecting an short package, which in my understanding is only the case if the transfer length is under 64 Byte? Also `rxpkts` is an unsigned integer so can never be smaller than 0, therefore I changed the check to equals.

With these changes QMK successfully enumerates with the HID descriptors.

### DIFF
--- a/os/hal/ports/RP/LLD/USBDv1/hal_usb_lld.c
+++ b/os/hal/ports/RP/LLD/USBDv1/hal_usb_lld.c
@@ -176,6 +176,9 @@ static uint16_t usb_buffer_next_offset(USBDriver *usbp, uint16_t size, bool is_d
  * @brief   Reset endpoint 0.
  */
 static void reset_ep0(USBDriver *usbp) {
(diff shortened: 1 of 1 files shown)


hanya:
Looks good for me. Thank you finding the problem in the driver.

KarlK90:
> Looks good for me. Thank you finding the problem in the driver.

Great! :slightly_smiling_face: 

@fpoussin feel free to merge on your approval!

fpoussin (approved):

fpoussin:
LGTM, thanks!

KarlK90:
Great! Thanks for the merge!

PR #319, opened 2022-04-09
[RP2040] Fix usb endpoint handling on RP2040

KarlK90:
After cross referencing the driver with the official tinyusb driver and the rp2040 datasheet, I noticed some deviations especially in the handling of reset, stall and setup packets.

Also the old behaviour was that endpoints have already been armed to receive or send usb packets after they have been initialized. This should only be done by the `usb_lld_start_in()` and `usb_lld_start_out()` functions that are invoked by the higher level usb driver functions.


KarlK90:
CC: @hanya

fpoussin (approved):

fpoussin:
LGTM :)

PR #322, opened 2022-05-10
STM32F411-DISCO board: fix pin assignments

trheinfels:
There were two problems with the previous pin assignments:

1. Due to a copy-and-past error, the pins 13 and 14 were mapped as AF for all ports instead of GPIOA (for SWD) only.

2. All free pins were assigned as analog inputs. This limits the maximum input voltage to +4V (see STM32F411xE datasheet, Table 11), making otherwise 5V-tolerant pins prone to latch-up.


fpoussin:
LGTM, thanks!

PR #328, opened 2022-07-17
[RP2040] Add RP2040 PWM driver

KarlK90:
All individual RP2040 PWM channels are mapped onto distinct ChibiOS PWM drivers, as this fits driver model which wants an independent timer per driver.


KarlK90:
@fpoussin ready for review now :slightly_smiling_face:

fpoussin (approved):

PR #331, opened 2022-08-19
Support for STM32F303xB based on STM32F303xC

plewka:
The xB variant is different in Flash size only - 128k only, not 256k


fpoussin (approved):

fpoussin:
Thanks!

sigprof:
Is this actually correct? The datasheet (https://www.st.com/resource/en/datasheet/stm32f303cb.pdf) says that the variants with 128K flash also have a different SRAM size (32K instead of 40K for xC chips).

STM32F303xB/STM32F303x datasheet, page 12

Unfortunately, I don't have any STM32F303xB chips to perform testing on actual hardware (and this might not really prove anything, because the situation with the SRAM size might be the same as with the flash size on STM32F103x8 chips — the extra memory might be present in the actual chips, but not guaranteed to work).

PR #335, opened 2022-09-17
[RP2040] fix i2c lockup with periodic tick enabled

KarlK90:
The system time in periodic tick mode can only advance if interrupts are enabled. As the disabling and aborting of the i2c peripheral was done in a critical section interrupts where disabled, resulting in a lockup as the timeout condition could never be exited. This patch fixes that by enabling interrupts for a short time during the busy waiting.


fpoussin (approved):

fpoussin:
Thanks!

PR #336, opened 2022-09-19
[RP2040] Add `vectors_base` and `end` symbol to RP2040 linker script

KarlK90:
...to allow copying the vector table to RAM on startup.


fpoussin (approved):

fpoussin:
Thanks again!

PR #339, opened 2022-10-02
[RP2040] explicitly disable NVIC irq on USB stop, explicit USB Bus connection and remove `USB_DEBUG` facilities

KarlK90:
The NVIC interrupt was still enabled after the USB peripheral was disabled and the USB driver object was de-initialized to default values. This resulted in USB IRQs still firing and subsequently de-referencing of invalid memory e.g. on SETUP packets which the handler still tried to handle.

Furthermore the USB bus was signaled a connected device at the end of `usb_lld_start` by pulling up `DP` which has to be done explicitly by the application using the `usbConnectBus()` macro.


fpoussin (approved):

KarlK90:
Thanks :slightly_smiling_face:

PR #345, opened 2022-10-29
Teensy 4.1: implement USB remote wakeup

stapelberg:
This allows waking up a computer from Suspend-to-RAM by pressing a key on the keyboard (with the QMK keyboard firmware, which uses ChibiOS for the Teensy).

related to https://github.com/kinx-project/kint/issues/59

related to https://github.com/qmk/qmk_firmware/issues/16934

### DIFF
--- a/ext/nxp-middleware-usb/device/usb_device_ehci.c
+++ b/ext/nxp-middleware-usb/device/usb_device_ehci.c
@@ -1555,6 +1555,13 @@ usb_status_t USB_DeviceEhciCancel(usb_device_controller_handle ehciHandle, uint8
     return kStatus_USB_Success;
 }
 
+static uint32_t ms_to_cycles(const uint32_t val) {
+  // 600 cycles at 0.6 cycles/ns == 1μs
+  const uint32_t cycles_per_us = 600;
+  const uint32_t ms_to_us = 1000;
+  return val * ms_to_us * cycles_per_us;
+}
+
 /*!
  * @brief Control the status of the selected item.
  *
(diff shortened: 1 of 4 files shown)


stapelberg:
I converted this PR to a draft for now, because I noticed that on a different PC, with this change, the keyboard seems to not work after suspend/resume (when the resume is not triggered by the keyboard). Will need to investigate.

stapelberg:
> I converted this PR to a draft for now, because I noticed that on a different PC, with this change, the keyboard seems to not work after suspend/resume (when the resume is not triggered by the keyboard). Will need to investigate.

Did some initial investigation: the symptom is that when the PC resumes from suspend-to-RAM, it resets all devices connected to my USB hub, and the keyboard does not enumerate any longer. 

Unfortunately, as soon as I compile in printf_debug in `lib/chibios-contrib/os/hal/ports/MIMXRT1062/MIMXRT1062/printf_debug.[ch]`, the problem no longer reproduces. I can replace the `printf_debug()` implementation with a 1ms sleep and the problem still doesn’t show up.

This indicates it’s either a missing delay somewhere, a race condition/timing issue, or adding extra sleeps papers over the issue. In any case — very annoying to figure out :( 

I’ll see if I can find some time to set up my NXP devkit to see if anything obvious shows up in a debugger.

stapelberg:
> I’ll see if I can find some time to set up my NXP devkit to see if anything obvious shows up in a debugger.

When running it on the NXP devkit, the issue doesn’t reproduce :(

Worse, in my current testing, waking the PC up via USB didn’t work anymore! 

I have a https://www.crowdsupply.com/great-scott-gadgets/luna ordered, so when/if it ships I might give this another look, but for now I’ll give up on this. I don’t have the right tools to effectively debug this.

stapelberg:
Actually, one thing came to mind: I think my hack to use `DWT->CYCCNT` directly instead of `hwTick` is not quite correct: `hwTick` is measured in ms, not in cycles. So, probably the USB stack delays are too short, which might explain why the extra delays paper over the issue.

I’ll try to give this a shot over the weekend

stapelberg:
> So, probably the USB stack delays are too short, which might explain why the extra delays paper over the issue.

I have updated the code to convert milliseconds to cycles to fix the delay timing, but unfortunately the “hanging USB enumeration” symptom still happens :(

PR #346, opened 2022-10-29
Teensy 3.6: do not restart USB stack after wakeup

stapelberg:
Without this change, the Teensy 3.6 (with QMK firmware) can wake a computer from Suspend-to-RAM, but the keyboard does not actually produce any keypresses until you un-plug and re-plug it.

A similar change was needed for the Teensy 4.x as well: https://github.com/ChibiOS/ChibiOS-Contrib/pull/345

related to https://github.com/kinx-project/kint/issues/59

related to https://github.com/qmk/qmk_firmware/issues/16934

### DIFF
--- a/os/hal/boards/PJRC_TEENSY_3_6/board.c
+++ b/os/hal/boards/PJRC_TEENSY_3_6/board.c
@@ -214,3 +214,9 @@ void __early_init(void) {
  */
 void boardInit(void) {
 }
+
+void restart_usb_driver(USBDriver *usbp) {
+    // Do nothing. Restarting the USB driver on the Teensy 3.6 breaks it,
+    // resulting in a keyboard which can wake up a PC from Suspend-to-RAM, but
+    // does not actually produce any keypresses until you un-plug and re-plug.
+}


stapelberg:
I’m getting the feeling that the `restart_usb_driver` function should be defined in QMK instead of in ChibiOS-Contrib (though there is precedent in `ChibiOS-Contrib/os/hal/boards/SN_SN32F240/board.c`!), but that probably requires switching from ChibiOS-Contrib’s `PJRC_TEENSY_3_6` board definition to creating a QMK-specific `IC_TEENSY_3_6` board definition…

fpoussin:
```
/home/runner/work/ChibiOS-Contrib/ChibiOS-Contrib/os/hal/boards/PJRC_TEENSY_3_6/board.c:218:25: error: unknown type name 'USBDriver'
```

Maybe a condition would do the job here?

stapelberg:
> Maybe a condition would do the job here?

Nah, I’m pretty certain now that the code in question belongs into QMK, as the `restart_usb_driver` function is only defined in QMK. I created https://github.com/qmk/qmk_firmware/pull/19077 to make that change.

PR #350, opened 2022-12-07
Fixed some issues with usb_msd for stm32l496

BMichiel:
Some info:
I'm working with Chibios21.11
I'm using the chibios-21.11x branch. (obviously)
I'm working on an STM32L496

When implementing a mass storage device I've had a few issues:
1. The return value of scsi_transport_receive did not seem right.
2. Implicit casting warnings in read_format_capacities of lib_scsi.c.
3. Since I'm using a 32bit processor msd_cbw_t and msd_csw_t are padded.
    sizeof(msd_cbw_t) does not match the length of the binary data received and
    sizeof(msd_csw_t) does not match the length of the binary data to send.

I've fixed point 3 by providing defines for the size, since these messages should always
have the same length.

Comments are appreciated if this is considered an 'ugly' fix.

### DIFF
--- a/os/hal/include/hal_usb_msd.h
+++ b/os/hal/include/hal_usb_msd.h
@@ -98,6 +98,12 @@ typedef struct {
 } msd_cbw_t;
 
 /**
+ * @brief The size of the command block wrapper structure.
+ * @details The actual size of the structure ignoring data padding.
+*/
+#define USB_MSD_CBW_T_SIZE 31
+
+/**
  * @brief   Represents command status wrapper structure.
  * @details See USB Mass Storage Class Specification.
  */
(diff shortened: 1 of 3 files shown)


fpoussin (dismissed):

fpoussin:
maybe you could use the [_packed_](https://gcc.gnu.org/onlinedocs/gcc-10.4.0/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes) gcc attribute to remove padding?

BMichiel:
Sounds like a more fitting solution. I'll try that. (I'll be away during the holidays though.)

BMichiel:
Nice, in retrospect my change looked a little silly. Nice suggestion!

Tested: I can still mount the emmc as MSD over USB.

fpoussin (approved):

fpoussin:
Thank you :)

PR #352, opened 2022-12-09
[RP2040] update i2c drivers to reflect peripheral number

KarlK90:
RP2040 i2c peripherals are numbered starting from zero, therefore the drivers shall follow the same numbering e.g. I2C0 uses the driver I2CD0. This commit fixes the current miss match.


fpoussin (approved):

fpoussin:
LGTM, thanks!

PR #353, opened 2022-12-14
HT32: ld: Fix vector table alignment for HT32F523xx

hansemro:
For context, the page size of HT32F523xx is 512 bytes (see page 42 of HT32F52342/HT3252352 User Manual v1.30) and certain bootloaders jump to addresses that are not aligned to 1024 bytes (specified by common rules_code.ld). This commit updates the example ld scripts so that these bootloaders can correctly map the vector table and boot a ChibiOS application firmware.


fpoussin (approved):

fpoussin:
Thanks!

PR #355, opened 2022-12-31
RP2040: Init buffer position in adc_lld_start_conversion()

sigprof:
The RP2040 ADC driver was initializing `adcp->current_buffer_position` and friends only in `adc_lld_start()`, therefore a simple `adcConvert()` call did not work when called a second time (only the circular mode worked, because the interrupt handler was resetting the position in that case after the whole buffer was filled).  Reset the buffer position variables in `adc_lld_start_conversion()` to make the one-time conversion work.

This fix is needed to make the ADC support in QMK work (the QMK code does not use any fancy modes except blocking `adcConvert()` with a single channel and a single sample).


JohnAZoidberg:
Can confirm that it works. Tested on Rpi Pico.

fpoussin (approved):

fpoussin:
Thank you!

PR #361, opened 2023-01-30
HT32: SPI: avoid write collisions when SPI FIFO is disabled

hansemro:
This improves SPI transmission reliability when SPI FIFO is not used. Without this change, I experienced hangups when frequently transmitting data to SPI flash.

For reference:
When SPI FIFO is enabled, a write collision occurs when writing to SPIDR register while both the TX buffer and SPI shift register are full (which is when TXE flag is not asserted).

When SPI FIFO is disabled, a write collision occurs when writing to SPIDR register while both TX FIFO and TX shift register are full. Here, the TXBE flag remains appropriate since it is used to check when TX buffer is empty and when the TX FIFO is below a set threshold value.


fpoussin (approved):

fpoussin:
Thank you!

Issue #363, opened 2023-02-13
[STM32F37x] Using two comparators simultaneously bug


integernick:
I've recently run into a bug while trying to use both of the analog comparators on STM32F37. 
The problem seems to be that on STM32F37 and STM32F38 COMP1 and COMP2 share the same 32-bit register COMP_CSR, so configuring the second comparator affects the configuration of the first one.

Here's an example:

![image](https://user-images.githubusercontent.com/68876760/218428536-0821e1f4-b7e6-4b69-a352-b16fb03addd5.png)
![image](https://user-images.githubusercontent.com/68876760/218428428-36e80c6b-db7b-41d1-a7d0-c254c9203041.png)
*Breakpoint 1: COMP1 configured.*

![image](https://user-images.githubusercontent.com/68876760/218428499-d6454724-908c-4e4c-bcef-dfc2887fe113.png)
*Breakpoint 3: Enabling COMP2 discards the COMP1 configuration.*

I was able to fix this by adding some preprocessor directives to the `os/hal/ports/STM32/LLD/COMPv1/hal_comp_lld.c`, will create a pull request later.

ChibiOS version: stable_20.3.x
Compiler: arm-none-eabi-gcc (10.3-2021.10)
Platform and board: STM32F373CC (custom board)

PR #364, opened 2023-02-13
stm32f37x: Fix simultaneous comparator usage (#363)

integernick:
Fix simultaneous COMP1 and COMP2 usage on STM32F37x and STM32F38x.


fpoussin (approved):

fpoussin:
Nice catch, thanks!

PR #365, opened 2023-02-23
[RP2040] Fix host remote wakeup

KarlK90:
Previously a device which attempted to wake a sleeping host would not be notified about the resumed operation and thus stay suspended for ever. This is due to the fact that only a host initiated wakeup would trigger a interrupt. As a workaround we use the SOF interrupt which is enabled for a short amount of time to detect the resume of the bus and wakeup the device.


fpoussin (approved):

fpoussin:
Thank you!

PR #366, opened 2023-03-22
Add a generic board for lpc11u35

hftsai256:
Introduce a new generic board targeting lpc11u35 MCU as a foundation for applications like QMK. Also, several internal registers are renamed to avoid conflicting against some GPIO shorthand such as `B0` and `B1`.


fpoussin (approved):

fpoussin:
Thank you

PR #372, opened 2023-07-02
Fix unused parameter in EEPROM driver

elagil:
GCC fails to build for me, because I have warnings as errors, and warnings for unused parameters.
This fixes the build of the EEPROM driver module.


fpoussin (approved):

fpoussin:
Thank you!

PR #381, opened 2023-10-31
W5500 ethernet driver

xaelsouth:
W5500 ethernet driver for ChibiOS. W5500 is initialized in in Ethernet Mode and requires LWIP to communicate. No built-in TCP/IP stack is used.


fpoussin:
is mbedtls required for a simple example?

xaelsouth:
mbedtls ist not really neccessary, but nice to have, because it can be seen how mbedtls can be used.

fpoussin:
Could you strip it down to the bare minimum for a working tcp/ip example?
It's really big as of now.

PR #382, opened 2023-11-23
[RP2040] i2c: fix speed calculation

KarlK90:
The former speed calculation where way off the requested i2c speeds. This fix, which is based on the official pico-sdk implementation fixes the calculation for the high and low times:

 | Speed | Before | After |
 | ------| ------ | ----- |
 | 100k  | 126k   | 94k   |
 | 400k  | 660k   | 346k  |
 | 1000k | - (*)  | 760k  |

(*) Failed to start any transaction in my testing setup.


fpoussin:
Thank you!

PR #383, opened 2023-11-25
[RP2040] usb: fix usb_lld_get_status functions

KarlK90:
This fixes the USB endpoint get status functions; previously the (wrong) assumptions were the following:

* `active` was toggled depending on whether a transaction was currently ongoing on the endpoint e.g., if the endpoint is "busy." What is actually requested is if this endpoint is enabled at all. Therefore, we check the `enabled` flag in the endpoint control register.
* `stalled` was toggled by the device itself, although this property is completely controlled by the host via the SET and CLEAR feature requests. Therefore, we check the `stalled` flag in the endpoint buffer control register.


fpoussin:
Thanks

PR #385, opened 2023-12-29
SN32: implement EFL LLD for flash access

dexter93:
Get rid of BSP refs for FLASH access, implement ChibiOS EFL


fpoussin (approved):

fpoussin:
Very nice, thank you!

PR #386, opened 2024-01-19
SN32: I2C LLD makeover

dexter93:
It actually works now!

Extended for the second I2C hardware available in the family


fpoussin (approved):

fpoussin:
Thanks!

PR #388, opened 2024-01-23
Community I2C Fallback extended driver

dexter93:
Extends support for actively driving I2C via GPIO push-pull instead of passively through open-drains.
Useful for chips not having open-drain support in the GPIO peripheral.

Defaults to open-drain, selectable by `SW_I2C_USE_OPENDRAIN`

Compared to the default I2C fallback driver, this one also has further delays added to conform with the I2C protocol where appropriate, should be more stable.
Tested extensively in both open-drain and push-pull config. Both standard and fast mode I2C is achievable reliably

### DIFF
--- a/os/hal/lib/fallback/I2C/hal_i2c_lld.c
+++ b/os/hal/lib/fallback/I2C/hal_i2c_lld.c
@@ -1,0 +1,482 @@
+/*
+    ChibiOS - Copyright (C) 2006..2018 Giovanni Di Sirio
+
+    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


dexter93:
Currently simply replacing the Chibios i2c fallback driver call in `os/hal/ports/{PLATFORM}/LLD/I2C/driver.mk`. Support for it offloaded to platform maintainers, example usage in the SN32 family. Let me know if there's a better way to include this.

dexter93:
Closes #389

fpoussin (approved):

fpoussin:
All good, thanks!

PR #393, opened 2024-02-20
NUC123: add GPT timer LLD

hansemro:
This introduces a LLD for the 4-channel 24-bit timers available for NUC123 with support for continuous and oneshot modes.

This LLD does not provide support event counter mode for external counting pin function.


hansemro:
Added 24-bit compare value mask and updated comments.

hansemro:
Updated to better follow style guide (two space indents, CRLF line endings, etc.). Lengthy lines do not exceed 100 columns.

fpoussin (approved):

fpoussin:
Thanks!

PR #410, opened 2024-09-11
SN32 USB LLD: flexibility updates

dexter93:
Refactor the USB LLD to be compatible with the whole sn32f2 family of chips


fpoussin:
Very nice, thank you!

PR #411, opened 2024-09-11
SN32 Init cleanup

dexter93:
Some very specific options related to QMK and specific board setup were leftover here.
Remove them for generic chibios boards


fpoussin:
Thanks!

PR #412, opened 2024-09-18
add support arterytek at32f405

ArteryTek:
add support arterytek at32f405.


fpoussin:
That's great, thank you very much.

PR #413, opened 2024-12-04
Another update for AT32F415 

HorrorTroll:
### Changelog:
- Update board files to able use SDIO pin by default.
- Added TMR alias for ISR and small fix HAL LLD source files.
- Small fix for GPIO and SDIO LLD drivers.
- Update demo board and HAL tests.
- Added support CRC LLD driver.
- Fix the issues where CRC now only use software file if hardware not available,
- Added halconf_community for demo board and HAL tests.
- Fixed OTG driver does not re-enables endpoints on wakeup and other.


fpoussin (approved):

fpoussin:
LGTM, thanks!

PR #414, opened 2025-01-05
In LPC port, alias PUSHPULL to default digital output

anonymouscowhead:
LPC11Uxx does not support true pushpull output, but its non-opendrain output is sufficient in many use cases.
It's convenient to alias PUSHPULL to the default digital output, rather than simply undefining it here and leaving it undefined:  This avoids re-working application code that calls for PAL_MODE_OUTPUT_PUSHPULL.


fpoussin (approved):

fpoussin:
LGTM, thanks!

PR #418, opened 2025-02-14
lib_scsi: configurable size of read/write buffer

rusefillc:
performance optimization gives us 30% improvement on SPI SD card reads


dron0gus:
30% boost is archived by increasing read/write buffer from 512 bytes to 2048 bytes.

fpoussin:
Could you update the docs and whichever files are using the driver with the new field?

PR #424, opened 2025-06-08
stm32H7 compatibility and fixes to header

dynfer:
Fix LTDC clock init to be the same across the H7 lineup.
Fix the header to be usable from CPP, the current enum forward breaks compatibility.


fpoussin (approved):

fpoussin:
LGTM, thank you!

PR #435, opened 2026-02-22
fix: WB32 DMA Fixes

emolitor:
* Add validation for DMA Block Transfer Size Mask and use constant instead of hard coded value
* Fix ISR channel initialisation to only initialise channels that are in use
* Fix DMAC1 reset which was guarded with wrong channel
* Fix DST_AUTO_RELOAD_DIS which was off by one


fpoussin:
Thank you!

PR #436, opened 2026-02-22
fix: WB32 Timer Fixes

emolitor:
* Remove TIM8 and TIM20 which do not exist on this hardware
* Add guard for BDTR as only TIM1 is an Advanced Timer
* Fix off by one error OC5M and OC6M


fpoussin:
Thank you!

PR #437, opened 2026-03-15
Add Support for CH579M

ryanm101:
This is a port of the CH579M chip as i have a cheap macropad with that chip that does not support qmk. 

This port is a pre-req QMK.

Notes: 

This is heavily AI (Claude) assisted as this is not an area i am familiar with. I've marked WIP as I'm still working on the QMK part and there may be further tweaks needed at time of PR QMK Compiles, and is recognised by the host system, and qmk console can connect to keyboard.

This has enough now that QMK can build and work. There might be other features people want e.g. BLE etc but i dont have a board with that so cant test and i have scrubbed this boards Analog keys so i am now limited in what i can do without bodge wires.

### DIFF
--- a/os/common/startup/ARMCMx/compilers/GCC/mk/startup_ch579m.mk
+++ b/os/common/startup/ARMCMx/compilers/GCC/mk/startup_ch579m.mk
@@ -1,0 +1,30 @@
+# CH579M startup files (ARM Cortex-M0, ARMv6-M).
+#
+# Uses the generic ChibiOS ARMv6-M startup (crt0_v6m.S + vectors.S)
+# with WCH's CH579M CMSIS device header for the vector table layout.
+#
+# The system_ch579m.c provides SystemInit() which is called from crt0;
+# our hal_lld_init() then does the PLL switch in __early_init().
+
+STARTUPSRC = $(CHIBIOS)/os/common/startup/ARMCMx/compilers/GCC/crt1.c
+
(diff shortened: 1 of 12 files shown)


fpoussin:
Thank you!

os/common/startup/ARMCMx/compilers/GCC/mk/startup_ch579m.mk
```diff
@@ -0,0 +1,30 @@
...
+             $(CHIBIOS)/os/common/startup/ARMCMx/compilers/GCC/ld \
+             $(CHIBIOS)/os/common/ext/ARM/CMSIS/Core/Include \
+             $(CHIBIOS_CONTRIB)/os/common/startup/ARMCMx/devices/CH579 \
+             $(CHIBIOS_CONTRIB)/os/hal/ports/WCH/CH579 \
+             $(KEYBOARD_PATH_1)/ld
+
+STARTUPLD         = $(CHIBIOS)/os/common/startup/ARMCMx/compilers/GCC/ld
+STARTUPLD_CONTRIB = $(KEYBOARD_PATH_1)/ld
+
+LDFLAGS += -L$(KEYBOARD_PATH_1)/ld
```
zvecr, line 25: `KEYBOARD_PATH_1` is a QMK specific variable, and should not have been included in this lower layer dependency.

ryanm101:
Should I just submit a patch or would you prefer revert and resubmit?

